PR #733, opened 2013-11-29
add parallax horizontal and vertical properties to enable non-calculated parallax offsets

lukekarrys:
This enables users to specify the amount that the parallax background will be offset for each slide change. This is especially helpful if the parallax background is an image that can be repeated horizontally.

This also allows for only one direction of parallaxing to occur (such as horizontal but not vertical) by setting that property to 0.


hakimel:
Merged with some tweaks.

Sorry about taking forever to review this.

lukekarrys:
Thanks @hakimel!

PR #750, opened 2013-12-09
Added slide deck logo feature to enable displaying a fixed logo image on...

pepl:
If you would like to display a logo image on every slide, you can do so by setting `slidedeckLogo` configuration value to the URL of the image.

screenshot from 2013-12-09 22 40 37


handsomeyang:
Check out my example https://github.com/handsomeyang/reveal.js

PR #775, opened 2014-01-10
added reverse fragments (exits)

karimsa:
I added a `.reverse` class for `.fragment` elements which need to only go in reverse (a fragment which only exits). I came across this need while making a presentation.

Works like this:

```html
<div class="fragment reverse">I am a fragment</div>
<div class="fragment roll-in reverse">I am a fragment</div>
```


hakimel:
No thanks. I'd rather add new fragment styles than reversing the existing ones.

PR #819, opened 2014-03-10
Added option to force vertical parallax effect

TobbenTM:
Implemented option to force vertical scrolling of background when there are no vertical navigation in presentation.

Thought it might be handy. :)

Created a quick demo here: http://tobbentm.com/5f5/

Let me know what you think.


hakimel:
Thanks for the PR but I prefer sticking to just the existing type of parallax effect.

TobbenTM:
No problem. :)

PR #840, opened 2014-03-19
Add support for Live Reload

aseigneurin:
When preparing a presentation, it's quite handy to automatically reload the view in the browser when the content changes. I've enabled that with the Live Reload capability of the grunt-watch plugin (don't forget to run 'npm install' again).

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -97,6 +97,9 @@ module.exports = function(grunt) {
 		},
 
 		watch: {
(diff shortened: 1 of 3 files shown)


hakimel:
Thanks for the PR, live reload has been merged and enabled through an earlier request #720

Sorry about the delayed reply on this.

finid:
:+1:

PR #854, opened 2014-03-27
Make revealJS screen reader friendly by announcing the contents of each ...

mpnkhan:
Persons with Visual disabilities use Screen readers and currently none of the contents are announced and couldn't navigate with arrow keys. 

Please make this presentation framework accessible for Keyboard and Screen reader users.


marcysutton:
Just a heads up that `role="application"` and the use of an `aria-live` region in Reveal is actually making accessibility worse in JAWS and IE, the most used browser/AT combination. Because the text content is duplicated in the page for `aria-live`, the content of each slide is read more than once (3 times, actually). I'm working on a pull request to make this better, but I think focus management is probably a better way to go.

vick08:
Please "no", let's talk first. I need live regions when presenting live!

PR #917, opened 2014-05-16
Merge of Reveal.js With FontAwesome Framework

crood:
made the Fontawesome Framework fit to suit into your reveal.js Framework.
Thought its a nice way of having scaleable Icons already at hand, and they suit the style of the general reveal themes very well.
Added on the second slide a rotating spaceship as quick example that it works. (included those in the index.html) Needs to be downloaded to view tho, since the projectlink goes to the original reveal.js example.

(Not sure how to add the FontAwesome Guys into this pull request.)

Cheers, Crood


crood:
Alright ! :)

PR #928, opened 2014-05-29
Update Themes Slide

ryanplasma:
Added 'Blood' theme and fixed 'Default' theme if the current theme is not default.


hakimel:
The 'blood' theme is now included on that slide.

PR #940, opened 2014-06-05
notes: reset timer on 'r' key press.

mdebski:
I used reveal.js together with notes plugin and was annoyed that timer started on opening notes.
Usually you first prepare presentation, together with notes window, and spend a moment before starting it.
So I think it's desirable to have a functionality to reset timer.
Another use case is when a few people speak one by one, with common shared slides, and would like to measure time separately.

Please review the change, I may improve it if needed!

### DIFF
--- a/plugin/notes/notes.html
+++ b/plugin/notes/notes.html
@@ -180,8 +180,20 @@
 						nextSlide = document.getElementById( 'next-slide' ),
 						silenced = false;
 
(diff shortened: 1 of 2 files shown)


hakimel:
It's now possible to reset the speaker notes timer by clicking on it.

Thanks for the PR and sorry about the late reply!

PR #951, opened 2014-06-18
Feature prevent swipe

Calyhre:
Add ability to prevent swipe for child elements with the `prevent-swipe` attribute.

Basically, it allow users to add scroll/swipe interaction on elements, without swiping to another slide.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -3482,6 +3482,8 @@
 	 */
 	function onTouchStart( event ) {
 
+		if(preventSwipe(event.target)) return true;
+
 		touch.startX = event.touches[0].clientX;
 		touch.startY = event.touches[0].clientY;
 		touch.startCount = event.touches.length;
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks! This is really useful. Merged with two minor changes:
- The attribute was renamed to `data-prevent-swipe` to match other reveal.js arguments which are prefixed with "data-".
- The internal method was renamed from `preventSwipe` to `isSwipePrevented`. The first name sounded like the method itself was preventing swipes when it was just checking.

Calyhre:
:+1:

maxmeyer:
Is there any documentation for this feature?

hakimel:
@maxmeyer Good catch – it wasn't but I just added it in a151df54be33894a4b06c3a0bc9b51524db77e2c.

maxmeyer:
Cool. Thanks. :-)

PR #963, opened 2014-07-07
Display speaker notes on PDF print-outs

emmajane:
When creating a PDF of my deck I wanted the speaker notes to be available as well. I added the following CSS to `css/print/pdf.css`:

       191 .reveal aside.notes {
       192   border-top: 1px solid black;
       193   display: block !important;
       194   font-size: 16pt;
       195   height: 100%;
       196   line-height: 1.2em;
       197   margin-top: 60px;
       198   text-align: left;
       199 }

It worked well for my needs except on one or two slides where the region wasn't quite tall enough. I tried playing with the `section` visibility, but this didn't seem to make a difference. Suggestions on styles which would make all speaker notes display would be much appreciated (or if I'm approaching this the wrong way, suggestions on how to do this).

Slides with a problem:
clipped
clipped-2

Slides without a problem:
notclipped-2
notclipped

### DIFF
--- a/css/print/pdf.css
+++ b/css/print/pdf.css
@@ -52,14 +52,15 @@ html {
 /* SECTION 3: Set body font face, size, and color.
    Consider using a serif font for readability. */
 body, p, td, li, div {
-	font-size: 18pt;
+	font-size: 16pt !important;
 }
 
 /* SECTION 4: Set heading font face, sizes, and color.
    Differentiate your headings from your body text.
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks for the PR and sorry about the super late reply. 

I took a different direction with this, there's now a `showNotes` config value which makes it possible show notes on top of the deck both online and in PDF exports. More info: https://github.com/hakimel/reveal.js/blob/master/README.md#share-and-print-speaker-notes

PR #1009, opened 2014-09-12
CSS-based vertical centreing - makes layout JS free

burnpanck:
The centreing is based on a wrapper ```div``` which can be centred on the slide using CSS techniques. That wrapper is created during presentation initialisation in the ```setupDOM``` JS routine.

Furthermore, now the slides (i.e. the ```section``` CSS blocks) are fixed-size and well placed. This brings a clear distinction between placement/layout of elements on the slide and placement of the slide in the presentation universe (slide transitions/overview). It opens the path to more complex slide layouts, where e.g. the title is at the top but the rest of the contents are vertically centred, or slide footers, etc. 

This could possibly break user's ```section > some-content``` selectors, since there will be an intermediate div that is not visible to the user during the creation of the presentation. Such selectors should be replaced by ```.slide-content``` selectors, that class is added to the wrapper ```div``` just for that reason.


hakimel:
Thanks for the PR but I don't want to add the additional wrapper inside of each slide so this won't be merged.

I wish there was a way to vertically center without additional wrappers... Have been experimenting with flexbox but even then I've been unable to make it work, due to the vertical slides and scaling of slides.

burnpanck:
How about HTML5 custom elements / web components:
```html
<section>
  <slide class="center">
    <title>...</title> <!-- or div class="title" -->
    ... content ...
  </slide>
  <slide title="...">...</slide>
</section>
```
Then, the wrapper div's would be hidden behind the shadow DOM of the custom element, where css centreing would be possible, without having to explicitly modify the DOM. Particularly, users could apply custom CSS to the contents without having to worry about the slide arrangement DOM elements. Furthermore, the complete theming CSS could be applied scoped to the shadow DOM, possibly relieving us from `.reveal` all over the theme CSS.
I see great value in this, I'm almost tempted to give it a shot...

Issue #751, opened 2013-12-10
Fullscreen video/iframe


uhaveamsg:
Is this possible, i applied the class="stretch" but it seems to be still not taking the full width & height of the slide.

laishi:
I use 
<section>
<iframe frameborder=0 width=95% height=586 marginheight=0 marginwidth=0 scrolling=no src="pages/music.html"></iframe>
</section>

give me  width 960 ,  I want width 95% in my page,how can fix it!

hakimel:
Closing. This is in master since 3.0 was released.

amenk:
You mentioned it is in master, but was is the way now to do an iframe as background but with controlls enabled? Can't seem to find it in the docs. data-src is mentioned to work only without being able to controll the iframe contents. And `class="stretch"` does not use the full screen, but only the max slide content width.

nevesnunes:
If anyone wants background video with controls, you can apply this workaround:

```js
(new MutationObserver(function(mutations, observer) {
    if (mutations[0].addedNodes.length) {
        mutations[0].addedNodes.forEach(function(el) {
            if(/video/i.test(el.tagName)) {
                el.controls = true;
            }
        });
    }
})).observe(document.body, {
    childList: true,
    subtree: true
});
```

PR #1029, opened 2014-10-09
Add Iframe background support

lutangar:
Hi there,

Regarding #751 , I made few changes to handle an `Iframe` as  a background through the `data-background-iframe` attribute:

```
<section data-background-iframe="https://github.com/hakimel/reveal.js">
</section>
```

This way you can use any webpages as a background.

Cheers,
Lutangar.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -672,6 +672,7 @@
 			backgroundSize: slide.getAttribute( 'data-background-size' ),
 			backgroundImage: slide.getAttribute( 'data-background-image' ),
 			backgroundVideo: slide.getAttribute( 'data-background-video' ),
+			backgroundIframe: slide.getAttribute( 'data-background-iframe' ),
 			backgroundColor: slide.getAttribute( 'data-background-color' ),
 			backgroundRepeat: slide.getAttribute( 'data-background-repeat' ),
 			backgroundPosition: slide.getAttribute( 'data-background-position' ),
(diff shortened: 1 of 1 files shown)


justmytwospence:
+1 to @AlJohri's suggestion to adjust the z-index for iframe's exclusively

gevero:
@AlJohri The solution seems good enough but presents two problems:
* When data are loaded, it works only for the first display of the slide, then if you go back the z-index stays to 0. In order to work one could add the following lines after the `data-loaded` if block:
```
if( background.hasAttribute( 'data-loaded' ) === false ) {
            // ...
            if( backgroundImage ) {
                // ...
            }
            // ...
           else if ( backgroundIframe ) {
               // ...
              document.querySelector('.reveal > .backgrounds').style['z-index'] = 1;
           }
}
if (slide.getAttribute( 'data-background-iframe' )) {
     document.querySelector('.reveal > .backgrounds').style['z-index'] = 1;
}
```
* If viewDistance>1 the z-index settings in changed also for the other loaded slides, since the the changed setting is presentation wide, therefore in order for this hack to work one needs to set viewDistance=1.

sumitrshah:
@AlJohri  Instead of changing the Reveal library, I think the best way to implement this hack is to listen for the 'slidechanged' event and then change the backgrounds z-index if the current slide uses an iFrame background.

For example, if using Reveal in combination with jQuery, I would use the following to make background iFrames reactive:

```JavaScript
Reveal.addEventListener('slidechanged', function(event) {			  	
     if($(event.currentSlide).attr('data-background-iframe')) {
          $('.reveal > .backgrounds').css('z-index', 1);
     } else {
          $('.reveal > .backgrounds').css('z-index', 0);
     }				    
});
```

Maybe @hakimel will add an option to making iFrame backgrounds active.

Issue #1036, opened 2014-10-24
Background video starts to play in speaker notes window


Suva:
I am not sure if it is a current bug, I am using slightly older version of reveal.js dev branch. I have two slides in front of one with background video. When the next slide has video, it will automatically start playing it in next slide preview, with sound.

Is there some workaround for disabling the video in speaker notes or at least mute it?

PR #1037, opened 2014-10-24
Do not add video backgrounds to speaker notes.

Suva:
Do not add video backgrounds in speaker notes window as per issue #1036

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -2601,7 +2601,7 @@
 					background.style.backgroundImage = 'url('+ backgroundImage +')';
 				}
 				// Videos
-				else if ( backgroundVideo ) {
+				else if ( backgroundVideo && !isSpeakerNotes() ) {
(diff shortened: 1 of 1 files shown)


jeertmans:
Hello @hakimel, sorry to re-open the discussion, but wouldn't it be better to mute the audio?

I have background-video only slides, and not showing them in the speaker notes makes the view pretty useless. To me, playing videos but muted would make much more sense.

drsybren:
> I have background-video only slides, and not showing them in the speaker notes makes the view pretty useless. To me, playing videos but muted would make much more sense.

I was actually considering writing a bug report that videos don't show in speaker notes, but then I found this report.

drsybren:
You're on fire, thanks!

jeertmans:
Hello @hakimel, do you already know when this will be part of a new release?

PR #1054, opened 2014-11-14
A new plugin: footnotes

niedakh:
Hi,

I've implemented a footnotes & bibliography plugin for reveal.js, please consider pulling it in. An explanation of how it works is in plugins/footnotes/README.md

The idea is that sometimes you need to put fixed positioned footnotes but let their content change slide by slide, especially in scientific presentations. You also then need a slide with bibliography at the end. So I've implemented this. Only thing I am not sure how to resolve yet is a bibliography longer than a sensible count, let's say 10. Perhaps by enabling scrolling inside the slide, will need to think of it, but that's easy to style by oneself if needed. Please comment, I am willing to develop this further/fix it, as I will be using this really often.

Note: I've regenerated the .css files, this makes the diffs a little less readable, but if you skip the .css files which were generated, you will see a clear and readable changeset. I've also added a test index.html file to test/ to show how all of this works and sensible (I think) defaults to the original file, which does not generate bibliography but exemplifies how footnotes can be used.

### DIFF
--- a/css/reveal.css
+++ b/css/reveal.css
@@ -8,7 +8,19 @@
 /*********************************************
  * RESET STYLES
  *********************************************/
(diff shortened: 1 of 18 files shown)


hakimel:
Hey! The reveal.js repository has started growing too large and hard to maintain so I have decided not to merge plugins directly into it anymore. The plan is to move existing plugins out of the repo too. More info: https://github.com/hakimel/reveal.js/wiki/Plugin-Guidelines

That said, I'm closing this but please feel free to add your plugin to the plugins list if you're open to maintaining it in your own repository: https://github.com/hakimel/reveal.js/wiki/Plugins,-Tools-and-Hardware

PR #1074, opened 2014-12-13
New theme: Diamonds

gnclmorais:
This theme is 95% based on simple.scss, the big change are the arrows.
I was never fond of reveal.js’s arrow, and always thought some kind
of squares would look muche better.

So, for the sake of checking how these themes work, I did this.
reveal.js doesn’t play along well with what I’ve done… That’s why
you can find a couple of !important on diamonds.scss.
Maybe a simple refactor would increase themes customisation…?


demoneaux:
@hakimel what do you think?

themar7777 (approved):

PR #1080, opened 2014-12-18
Enabled live-reload when changing index.html

nknapp:
I've added a change to the  Gruntfile so that editing index.html causes the browser page to reload (when `grunt connect watch` is running).


hakimel:
Thanks for the PR, live-reload has been enabled in #720

PR #1086, opened 2014-12-24
Fixed reload of slide with a custom ID containg special character

kosek:
In ID selector some reserved characters like ., #, {, }, ... must be
escaped. It's easier to use getElementById() method which doesn't require escaping.

The current reveal.js will not reload deck at specified slide if it has ID containing for example period, like <section id="R1.2.3"> - this will be turned into #R1.2.3 selector, which is invalid.

Some issues and complexities of escaping CSS identifiers are availabe here https://mathiasbynens.be/notes/css-escapes

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -2174,7 +2174,7 @@ var Reveal = (function(){
 		// assume that this is a named link
 		if( isNaN( parseInt( bits[0], 10 ) ) && name.length ) {
 			// Find the slide with the specified name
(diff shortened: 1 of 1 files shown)


hakimel:
This was fixed in 3.0 by validating link identifiers with regex, see 8c9c0ab0a64f32e8aa567009a82bac57464f4f93

I agree that `getElementById` is more appropriate so I went ahead and swapped that out in 817bb3bf43879009ae445e52d4b6ef88053c5583

PR #1097, opened 2015-01-09
Add Dependency badge in readme

PeterDaveHello:
Don't know if you'll like it :)


hakimel:
Thanks but I prefer not having that in the readme :)

PR #1102, opened 2015-01-14
base option is configurable in grunt connect task

adelura:
I'm using reveal.js core as a submodule in my project. This approach is nice because I don't have to store reveal.js core in my repo: https://github.com/CodeCarrotsJS/szkielet-prezentacji. But problem reveals when I want to run reveal.js server using `grunt serve` command in submodule directory(src/reveal.js). What I see is that `base` option is hardcoded to `.` I made it configurable to be able to run `grunt serve --serve-base=../..`.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,6 +1,7 @@
 /* global module:false */
 module.exports = function(grunt) {
 	var port = grunt.option('port') || 8000;
+	var base = grunt.option('serve-base') || '.';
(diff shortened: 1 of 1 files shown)


hakimel:
I was unable to merge this, looks like the originating branch was deleted. I've gone ahead and made the same change manually – though I named the option just "base" (rather than "serve-base") to match "port".

PR #1104, opened 2015-01-15
Feature: Show hidden fragments in speaker notes

fhemberger:
In my presentations, I work a lot with fragments to structure the narration during a talk. This PR highlights the hidden fragments inside the speaker notes, so you can see what fragment(s) will be revealed next.

I already used this patch on my own presentations for a while and it helped me a lot. I think others will benefit as well.


fhemberger:
@hakimel Any chance to get this merged?

hakimel:
Fragments are used to apply multiple different types of effects, including showing content, hiding content, making elements grow, shrink or change color. The styles in this PR seem specifically designed for fade-in fragments and would conflict with some of the other styles.

I'm not sure if it was there when you wrote this PR, but there's also a preview of the upcoming slide inside the speaker view. That shows you the upcoming fragment before it becomes visible. Perhaps that's enough of a preview of upcoming fragments?

If there's a lot of interest in showing all fragments in the speaker view I'd prefer making this a configurable option rather than changing the default behavior.

Sorry about never getting back to this. When I started reveal.js I thought I'd eventually "catch up" on all issues and PRs and reach inbox zero but I've long since realized that's never happening 🙈

fhemberger:
@hakimel Thanks for taking the time to review it!

PR #1106, opened 2015-01-18
Adding one-sided transitions

jammon:
Now the transition isn’t any more „per slide“ but „per movement of a slide“. So the slide can disappear in a different way than it appeared. This makes a more natural changing of the transition style during a talk.
There is an explanation added to README.md.

(This is my first pull request ever, so please be kind if I made any mistakes.)


hakimel:
This is really great! I'm creating some tests to ensure there's no regression but so far everything is working perfectly. Will merge soon.

hakimel:
Merged!

Should probably also add support for separate in and out background transitions (`data-background-transition`) at some point so its consistent.

jammon:
Great! Thanks for coming back on this!

lcrespom:
Would also be great if you could specify different in and out transition speeds

PR #1124, opened 2015-02-06
Add KaTeX math rendering plugin

j13z:
Hi,

I've added a math rendering plugin that uses [KaTeX](https://github.com/Khan/KaTeX), as an alternative to the existing plugin based on MathJax.

It renders all formulas in advance, so there is no need to re-render on `slidechange`. Feels much more lightweight I'd say. Supports the same `$…$` or `$$…$$` style replacements, as well as wrapping formulas in elements.

All changes were made to the plugin directory, except the tests, which I added to `test`. README and example included.

The plugin directory includes the KaTeX build, as I saw that the `markdown` plugin did the same with `marked`. 1,1 MB is quite heavy though, due to fonts (but runs offline then). If that's too big, KaTeX could perhaps be downloaded with a grunt taks or manually instead of adding it to the repo.


maxmeyer:
@hakimel 

Would it be possible to support an "plugin" api for external plugins and a reveal.js-contrib-repository to give all the wonderful plugins a home, but do not blow up the main repo to much?

hakimel:
Good idea for a plugin, thanks for the PR!

@maxmeyer Yeah, reveal.js absolutely needs to provide an easy way to
include plugins that live in other repositories. Merging all plugins
directly into this repo isn't scalable in terms of file size and, more
importantly, maintenance. Not sure what the right solution is yet.

j13z:
I wasn't really sure whether I should make an own repository for the plugin, or propose a pull request. If you don't want to blow up the repo size / maintain it, that's totally fine. I'll just create a dedicated repo for the plugin.

Anyway, I've added some changes to the PR, if it's still relevant:

I removed the KaTeX build (fonts and CSS, 1.1 MB) and added a CDN fallback. I however had to keep the JavaScript file for the tests, which is still quite heavy with 116 KB. (Maybe tests could use the CDN script too, but that would be slow and I didn't get it to work with headless QUnit yet).

plugin/math-katex/math-katex.js
```diff
@@ -0,0 +1,523 @@
...
+ * @author Johannes Schmitz
+ */
+window.RevealMath = window.RevealMath || (function() {
+	'use strict';
+
+	// --- Options and defaults ------------------------------------------------
+
+	var options = Reveal.getConfig().math || {};
+
+	// if (options.useCdn === true) {
```
d10: This option seems fine. Any reason why you commented it out?

plugin/math-katex/math-katex.js
```diff
@@ -0,0 +1,523 @@
...
+
+	function addStylesheet() {
+		// - To fix a CSS issue: Prevent reveal.js from overriding the font of
+		//   KaTeX elements.
+		//
+		// - To highlight errors (`.math-katex-error`)
+
+		var stylesheet = document.createElement( 'style' );
+
+		stylesheet.innerHTML  = '.reveal .katex { font-family: KaTeX_Main; }';
```
d10: You should do something like:
```js
var html = '';
html += '…';
stylesheet.innerHTML = HTML;
```

plugin/math-katex/math-katex.js
```diff
@@ -0,0 +1,523 @@
...
+			var s = e.innerHTML;
+			e.innerHTML = s.slice( 0, error.position ) +
+			              '<span class="math-katex-error">' +
+			              s.slice( error.position ) +
+			              "</span>";
+
+			// Log
+			console.error( 'Formula error on slide ' + slideNumber, error );
+
+			// Just show a `window.alert`
```
d10: IMO you shouldn’t create an alert as it would disrupt viewers. Logging it to the console seems fine to me.

plugin/math-katex/math-katex.js
```diff
@@ -0,0 +1,523 @@
...
+				});
+			}
+			catch ( error ) {
+				handleError( error, e );
+			}
+		});
+	}
+
+
+	// --- Script and stylesheet loading ---------------------------------------
```
d10: You could use `head.js`, since it is included in Reveal.js for loading of assets.

plugin/math-katex/math-katex.js
```diff
@@ -0,0 +1,523 @@
...
+
+
+	// --- Load KaTeX and run the plugin ---------------------------------------
+
+	// (Could really use promises. Avoided another dependency for now … )
+
+
+	/**
+	 * Loads KaTeX by first trying to load it locally (from `lib/katex`), then
+	 * from a CDN ass a fallback.
```
d10: Typo here: “ass” should be “as”.

d10:
Great work, @j13z! :smiley: Would love to see this in, perhaps you could even consider replacing the MathJax plugin! ;)

plugin/math-katex/math-katex.js
j13z: Oh thanks for the hint … I forgot to remove the comments …

This was refactored and the option is not needed anymore. The plugin automatically tries to load KaTeX locally and if that fails it makes a request to the CDN.

plugin/math-katex/math-katex.js
j13z: Nope, it *should* disrupt, otherwise you won't notice errors while editing. It will however only show one alert, for the first error.

d10: That's for the developer only, it would be a bad user experience to create an alert for something the user does not know.

plugin/math-katex/math-katex.js
j13z: Yeah, I’d like to get rid of that function as well, but if I’m not missing something then `head.load` would give me no error callbacks.

plugin/math-katex/math-katex.js
j13z: I hope so!

j13z:
@d10 Thanks. On replacing the MathJax plugin: That's probably not a good idea because KaTeX is not yet as "feature complete": E.g. MathJax can handle `\overset{}{}`, KaTeX can't.

@hakimel Please close this pull request, if you think the plugin should not go into the main repository.

plugin/math-katex/math-katex.js
j13z: I'm assuming here that you won't publish your slides with errors in formulas.

However, making this kind of "sideffect" switchable via an option would probably be a good idea for "embedded usage" like with http://slides.com/.

d10: Yeah, you should place it in an option which is disabled by default and add docs for it.

Sonaryr:
In my opinion reveal.js should remain "plugin free" and the plugins should live in their own repo. That way users can cherry pick the features they want. That's the reason I made my plugin in a different repo, because I think not everyone can benefit from it.

hakimel:
Yes, moving plugins to live in their own repos is the right way forward. Still considering options for how to best structure that and promote the available plugins from the main repo. There's currently a wiki page where some plugins are listed: https://github.com/hakimel/reveal.js/wiki/Plugins,-Tools-and-Hardware

That said, I'll close this PR since it won't be merged.

j13z:
I’ve created a repo for the plugin: [reveal.js-math-katex-plugin ](https://github.com/j13z/reveal.js-math-katex-plugin)

hakimel:
Great! Was gonna add it to the wiki but saw you already did that. I added a link to the plugin wiki page in the project readme as well.

maxmeyer:
@hakimel 

Should I open an issue for a discussion about a "good" api for plugins?

hakimel:
@maxmeyer Absolutely – feel free to.

I've started a "Plugin Guidelines" document which for now mostly states that plugins should live in separate repositories, and explains why. This document is also referenced from [CONTRIBUTING.md](https://github.com/hakimel/reveal.js/blob/master/CONTRIBUTING.md).

PR #1129, opened 2015-02-12
Contained deployment

2m:
This adds `grunt-cli` to `devDependencies` so it is not mandatory to install `grunt-cli` globally when running reveal.js.

Also this PR adds `--hostname` option to `grunt serve` command, so it is possible to set hostname to bind to.

With these two changes it becomes possible to deploy `reveal.js` to heroku by adding a `Procfile` with the following contents:

```
web: npm install && node_modules/.bin/grunt serve --hostname 0.0.0.0 --port $PORT
```

I've seen previous attempts (#801) of making reveal.js easier to deploy to heroku, and I apprieciate your goal of keeping reveal.js environment agnostic. IMO this change makes `reveal.js` self contained and allows for easier deployment to various environments.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -1,5 +1,6 @@
 /* global module:false */
 module.exports = function(grunt) {
+	var hostname = grunt.option('hostname') || 'localhost';
 	var port = grunt.option('port') || 8000;
 	// Project configuration
 	grunt.initConfig({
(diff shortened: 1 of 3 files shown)


bnjmnt4n:
We've since added `grunt-cli` to devDependencies to allow it to be used without global installation. I don't think adding the hostname option is a good idea as using Grunt as a wrapper to serve the code is not part of Grunt's scope of tasks, and there could be unexpected issues when doing so publicly. Instead you should use a specialised library for deployment.

2m:
Fair enough!

PR #1132, opened 2015-02-14
Add plain class for plain images

kendaleiv:
Add `plain` class for removing border and box-shadow from images.

Example usage:

```html
<img class="plain" src="some_url" />
```


Delapouite:
Yes! I've been using a similar `.raw` class for a while.

hakimel:
Makes sense, thanks!

Issue #1114, opened 2015-01-25
Code background and color should be defined by Highlight theme


borisschapira:
https://github.com/hakimel/reveal.js/blob/master/css/theme/template/theme.scss#L181

these lines are incompatible with several hljs themes like mono-blue ou github...

Is there a reason for these lines ? Could we remove them ?

aslakhellesoy:
I'm bitten by this as well - the highlight theme's background and foreground colours are overridden by theme.scss.

borisschapira:
@hakimel, any explanation ? I would gladly make a Pull Request but I imagine you've made it that way for a reason.

shmerl:
Looks like it's the same bug as I stumbled upon in #1290.

demoneaux:
Closing this as a duplicate of #1290 which has been fixed.

PR #1135, opened 2015-02-16
Add option to override code colors

martin-helmich:
This commit extracts the background and base font color of code areas into SCSS variables, allowing them to be changed in custom themes.


Delapouite:
This should fix this issue : https://github.com/hakimel/reveal.js/issues/1114

martin-helmich:
This has been open for quite a while. Any feedback/thoughts on this?

hakimel:
Those styles were removed instead, so that you can use any highlight.js theme. 523f78eda724006e12a502a7b6d7c8147dc92fbe

PR #1138, opened 2015-02-19
Background video loop

denehyg:
This adds the 'loop' attribute to the video element for slide backgrounds when provided with a 'data-background-video-loop' attribute alongside the 'data-background-video' attribute. The background video will loop when it reaches the end - particularly useful for short, ambient background videos.


hakimel:
Merged, thanks!

PR #1141, opened 2015-02-19
Check before calling blur on activeElement.

jhofker:
It's possible for slides to be in a situation where the last clicked thing was an SVG before the tab/window loses focus. When returning, `.blur()` is called on the previously-active element, but can result in an exception.

This protects against that and will only call `.blur()` when `document.activeElement` supports it.


hakimel:
Thanks for fixing that.

This commit has been merged into the dev branch. In the future please submit all pull requests towards that branch. https://github.com/hakimel/reveal.js/commits/dev

Issue #1050, opened 2014-11-11
Markdown table column alignment


dnmfarrell:
Cell alignment in markdown is ignored:

    |ID|Name|Value|
    |------|:---:|-----:|
    |001|Bank ABC|5,740.00|
    |002|Super Fund|1,2280.00|
    |003|Big Savings Bank|430.00|

Should display as:

|ID|Name|Value|
|------|:---:|-----:|
|001|Bank ABC|5,740.00|
|002|Super Fund|1,2280.00|
|003|Big Savings Bank|430.00|

Instead you have to type this:

    |ID|Name|Value|
    |------|---|-----|
    |001|Bank ABC|5,740.00|
    |002|Super Fund|1,2280.00|
    |003|Big Savings Bank|430.00|

To get revealjs to display properly. The flexibility is helpful but this reduces the benefits of using a markdown approach as the syntax is ugly and verbose. This is a barrier in getting revealjs adopted at $work, particularly by non-technical users. 

Please just respect the markdown formatting!

Thanks

fehmer:
+1

GitHub-Lover:
+1

hakimel:
Fixed by @fehmer in #1149

dnmfarrell:
Thank you **very** much!

PR #1149, opened 2015-02-28
Markdown table column alignment #1050

fehmer:
The alignment information were already in the generated html.
The css was the only thing missing, so I added it.


hakimel:
Great, merged!

PR #1153, opened 2015-03-03
Multiplex

clemmen2:
This makes it easier on users to use multiplexing. It is as easy as ```node run client``` or ```node run clientSocket``` and ```node run master``` on the appropriate machines. All the user has to do is edit the _mpSet.json_ file with the appropriate values. 

This does use EJS instead of just html. The slides are edited using the _source/partials/slides.ejs_ file. No ejs magic in this file it is just so that the user does not have to worry about ```Reveal.initialize()``` (located in the _footer.ejs_ files). The _footer.ejs_ files have ejs magic that use _mpSet.json_ for the multiplexing parameters.


hakimel:
Thanks for the PR but this won't be merged. I don't want to switch to EJS and I don't think multiplexing scripts and config should be included in the default presentation. I'm sure others who'd like to use a more template-based approach for their presentations will find your fork useful though.

PR #1156, opened 2015-03-05
Add support for custom notes.html file

danieltdt:
It would be nice if we could define a custom notes.html file.

Actually, I'm used to compile my js files before releasing my app and, using selectors like `script[src$="notes.js"]`, doesn't work :(

So, what do you think about it?


danieltdt:
@hakimel any thoughts?

hakimel:
Sure, makes sense to make that more flexible. Thanks!

PR #1158, opened 2015-03-08
Fix slide numbering for custom slide number formatting.

bchretien:
Numbering was off when dealing with fragments. I simply split your code from `getProgess()` by adding a `getSlidePastCount()` (a better name could certainly be found), and use that to get the global flattened slide number. It would probably be better to only compute this index once (for both progress and slide numbering), but at least slide numbers are now correct.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -2476,11 +2476,12 @@
 			}
 
 			var totalSlides = getTotalSlides();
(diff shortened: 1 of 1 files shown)


hakimel:
Good fix, thanks!

bchretien:
You're welcome ;-)

PR #1162, opened 2015-03-14
Make npm dependency usable by Browserify through NPM

bobylito:
Right now the package is not recognized as usable library by Browserify.
Adding the main attribute in the package.json solves that, and now gives
others the possibility to use reveal as a library in more complex
systems.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "The HTML Presentation Framework",
   "homepage": "http://lab.hakim.se/reveal-js",
   "subdomain": "revealjs",
+  "main": "js/reveal.js",
   "scripts": {
     "test": "grunt test",
     "start": ""


hakimel:
Makes sense. Are there any side effects from this for non-Browserify users?

bobylito:
Checking the [npm doc](https://docs.npmjs.com/files/package.json#main), it doesn't seem to bring any side effects. It is just meta data so node/browserify/whatever knows which file contains the main script when you do a require.

hakimel:
Cool, just wanted to check. Merged!

Narigo:
This is exactly what I needed to use reveal.js via NPM. Will there be a new release to NPM, that includes this fix, soon? :)

Thanks!

Issue #1163, opened 2015-03-14
Enable SmartyPants option of Marked when rendering Markdown


callahad:
`marked.setOptions({smartypants: true});` should automatically change straight quotes like "" into nice quotes like &ldquo;&rdquo;.

I'd suggest this default to on, since Marked is smart and avoids applying these transformations to code blocks, but I'd be happy with it being more easily configurable, too.

Unfortunately, the version of Marked that's bundled with Reveal has a bug, but updating to the latest Marked release and setting the above option appears to work seamlessly.

hakimel:
I'm open to adding an option to the markdown plugin for this. Do you want to give it a shot? It would probably make sense to group markdown plugin options inside one object like:

```javascript
Reveal.initialize({
  markdown: {
    smartypants: true
  }
})
```

Which the markdown plugin then consumes via `Reveal.getConfig().markdown.smartypants`.

legends2k:
It would be good to have @hakimel suggestion, but till then putting this in `Reveal.initialize` under `index.html` works:

    { src: 'plugin/markdown/marked.js', async: false, callback: function() { marked.setOptions({smartypants: true}); } },

hakimel:
The proposed config option in the comments above was added by @demoneaux in #1765

PR #1165, opened 2015-03-16
Bump Marked to v0.3.3

callahad:
Compressed with `uglifyjs lib/marked.js --comments="/Copyright/" --mangle --compress`

Used Marked v0.3.3 as of chjj/marked@2b5802f258c5e23e48366f2377fbb4c807f47658

Mostly minor bugfixes and refactoring in the Marked changelog, though this revision *does* have working SmartyPants support (off by default), which makes my life easier in #1163.

It also closes an [XSS vulnerability on IE < 10](https://github.com/chjj/marked/issues/492) when `sanitize` is set to true (it defaults to off)... actually, there are evidently a bunch of bugs in Marked, including other XSS's, as discussed by the NodeBB folks in julianlam/nodebb-plugin-markdown#20

They ultimately settled on [remarkable](https://github.com/jonschlinkert/remarkable) as a replacement, which is much more actively maintained... and then the two primary authors of Remarkable forked it into markdown-it. Now there's [bad blood](https://github.com/jonschlinkert/remarkable/pull/130#issuecomment-71205884). So this is all kind of a mess. Markdown-it looks good.

Probably worth updating Marked in the interim, since it's a noninvasive update, then moving to Markdown-it when you/I have time for more extensive testing.


callahad:
Whoops, opened this against master instead of dev, one sec.

PR #1167, opened 2015-03-16
Attribute values should be quoted

drzax:


hakimel:
Thanks, this has been merged into the dev branch https://github.com/hakimel/reveal.js/commits/dev

Issue #1105, opened 2015-01-16
feature request: keyboard shortcut + API call to skip forward/back over fragments


jason-s:
The left and right arrow keys move forward and back in the presentation through each fragment at a time. If there are a lot of fragments, it is clumsy to go to the previous or next slide.

Could you add a feature to move forward and back, skipping over fragments? (e.g. shift-left arrow and shift-right arrow)

niccnacc:
I'd love to see that too.
https://github.com/shower/shower let's you present forwards in fragments and then locks the fragmets automatically upon showing the last one.
I guess a really clumpsy hack would be to use the "shown" callback of the last fragment and forbid the fragmentation somehow.
But imo this should be a general setting. I guess some people prefer permanent fragmentation and if the presentation is with question session it is really really bad to have it this way.

kellyk:
I created a pull request for this issue (though I wasn't sure how to link them):
https://github.com/hakimel/reveal.js/pull/1173

dsmf:
Would be awesome to have this feature. I just used RevealJS in order to prepare two presentations at work. Really love it except the navigation when there are fragments...

Lomacar:
The pull request was made 7 years ago. Why wasn't it merged?

BTTF2001:
> The pull request was made 7 years ago. Why wasn't it merged?

Sorry I'm new and didn't realise pls except my apologies

Lomacar:
The API has methods for `.next()`, `.right()`, `.down()` etc. And it has `.nextFragment()`. These all do the exact same thing right now, which seems useless. I guess the direction ones just simulate the key press but `.next()` and `.previous()` should jump slides and ignore fragments!

hakimel:
This has been added in 90bbe8be4f12ec56723a335d36f9be984ef7ed03

If you hold alt/option while pressing any navigation keyboard shortcut you'll skip all fragments. This works for the arrow keys as well as prev/next navigation via the space key.

To skip fragments in API navigations you'll need to include the "skipFragments" flag, for example: `Reveal.right({ skipFragments: true })`.

Also, if you hold shift while pressing an arrow key you'll jump to the last slide in the given direction.

PR #1173, opened 2015-03-21
Issue #1105: Keyboard shortcut to skip forward/back over fragments

kellyk:


niccnacc:
Looks really well implemented. Without having tried it, is there a way to automatically activate "hard change" commands locally, if the points were already shown? This would make it a production quality treat!
Thanks for your work!

bcokert:
Shift+right/left works as expected. Shift+up/down skips to the end of the fragments list, instead of to the next slide, which is a bit unexpected. Otherwise, works great! Used it in my last presentation (where I really just wanted let+right anyways).

bcouetil:
Hello, would love to have this feature, when is it expected to be merged ?

Lomacar:
PLEASE MERGE.

hakimel:
Very sorry about never getting back to you on this @kellyk. I'm perpetually behind on issues and PRs.

As of 90bbe8be4f12ec56723a335d36f9be984ef7ed03, holding alt + pressing any navigation shortcut (arrows, space, n, p etc) will now skip fragments. Shift+left/right were already used for jumping to the first/last slide we couldn't use that modifier. More info: https://github.com/hakimel/reveal.js/issues/1105#issuecomment-913565307

PR #1184, opened 2015-03-31
Add support for setup with Docker Container

lborguetti:
Hi Hakimel, 

Added support for running (setup) reveal.js in Docker container. I hope to be useful.

Thank you.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,10 @@
+FROM node
+
+COPY package.json /code/
+
+WORKDIR /code
+
+RUN npm install -g grunt-cli && \
+    npm install
+
+CMD grunt serve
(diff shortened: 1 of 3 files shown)


hadim:
+1

hakimel:
Thanks for the PR! I'd like to avoid having reveal.js cater to a specific way of deployment. I think it's best if the framework remains as environment agnostic as it can.

Maybe you can share your instructions in the [wiki](https://github.com/hakimel/reveal.js/wiki) instead? I'm sure others would find that useful.

lborguetti:
Hi @hakimel, not problem. I go put instruction about docker deployment in the Wiki.  Thanks.

PR #1188, opened 2015-04-02
Display .stretch images in overview mode. Fix #1187

### LINKED ISSUE #1187: Images with .stretch class are not visible in overview mode
In reveal.js 3.0.0, the following slide markup:

```
<section>
  <img class="stretch" data-src="https://s3.amazonaws.com/hakim-static/reveal-js/arrow.png">
</section>
```

when rendered in overview mode, reduces to a 1px x 1px image:

image

It renders fine as a regular slide, though.

sanand0:
The height of `.stretch` elements is computed by setting its height to 0 and comparing the height of the rest of the elements with the parent slide's height.

In Overview mode, the parent slide's height is set to `700px`. So setting the `.stretch` element's height to 0 has no impact on the parent's height. This results in the `.stretch` element _always having a zero height_.

This commit temporarily sets the parent slide's height to `auto` before computing the new height, and clears the height.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -1195,10 +1195,15 @@
 			// Change the .stretch element height to 0 in order find the height of all
 			// the other elements
 			element.style.height = '0px';
+			// In Overview mode, the parent (.slide) height is set of 700px.
+			// Restore it temporarily to its natural height.
+			element.parentNode.style.height = 'auto';
 			newHeight = height - element.parentNode.offsetHeight;
 
 			// Restore the old height, just in case
 			element.style.height = oldHeight + 'px';
+			// Clear the parent (.slide) height. .removeProperty works in IE9+
+			element.parentNode.style.removeProperty('height');
 
 			return newHeight;
 		}


sanand0:
**A note on browser compatibility**

`.removeProperty` is used to clear the `height` style. This works only in [IE9 and above](http://www.quirksmode.org/dom/w3c_css.html). If we need it to work in IE8, a simple fix from [aight.js](https://github.com/shawnbot/aight/blob/master/aight.js#L995) is available:

    (function(CSSSDProto) {

      function getAttribute(property) {
        return property.replace(/-[a-z]/g, function(bit) {
          return bit[1].toUpperCase();
        });
      }

      // patch CSSStyleDeclaration.prototype using IE8's methods
      if (typeof CSSSDProto.setAttribute !== "undefined") {
        CSSSDProto.setProperty = function(property, value) {
          return this.setAttribute(getAttribute(property), String(value) /*, important */ );
        };
        CSSSDProto.getPropertyValue = function(property) {
          return this.getAttribute(getAttribute(property)) || null;
        };
        CSSSDProto.removeProperty = function(property) {
          var value = this.getPropertyValue(property);
          this.removeAttribute(getAttribute(property));
          return value;
        };
      }
    })(CSSStyleDeclaration.prototype);

gw0:
Fixes #1187.

hakimel:
Looks good – thanks!

PR #1193, opened 2015-04-03
Added support for standard transforms

danielquinn:
-webkit- stuff only works on webkit browsers.  I'm pretty sure that using the standard `transform:` here works in most modern browsers.


Delapouite:
Here's your answer : http://caniuse.com/#search=transform

danielquinn:
Ah, well then I guess we don't even need the `-webkit-transform`?

hakimel:
Thanks

PR #1197, opened 2015-04-08
forwarded attributes are not correctly quoted

mazenovi:
causes bugs when tweaking data-background-size or data-background-position with values like "15% auto" or "top left"


hakimel:
Thanks but his was recently fixed in another PR #1167

Issue #702, opened 2013-11-12
MathJax and markdown


rajgoel:
Hello,

it appears that the MathJax plugin does not always work as expected when using markdown. I experienced that underscores and some brackets are not always properly processed, e.g

```
<section>
$\bar{X}= \frac{1}{T}\sum^T_{t=1} X_t, \bar{Y}=\frac{1}{T}\sum^T_{t=1} Y_t$
</section>

<section data-markdown  data-separator="^\n===\n$" data-vertical="^\n---\n$">
<script type="text/template">

$\bar{X}= \frac{1}{T}\sum^T_{t=1} X_t, \bar{Y}=\frac{1}{T}\sum^T_{t=1} Y_t$

===

$\bar{X}= \frac{1}{T}\sum^T&#95;{t=1} X&#95;t, \bar{Y}=\frac{1}{T}\sum^T&#95;{t=1} Y&#95;t$

</script>
</section>
```

All three slides should produce the same result, but in the second slide the formula is not processed - probably because the underscore is not provided to MathJax. I experienced similar issue with brackets. By including the HTML-code for the underscore the formula is produced correctly.

It would be great if the MathJax plugin would work without the HTML-code workaround when using markdown.

Thanks for developing such a great presentation software :-)

pedritomelenas:
Similar problems arise while using arrays or matrices. For instance, these two slides should look the same:

				<section>
					<h3>Mathjax support without markdow </h3>

					$$
					\begin{pmatrix}
					1 & 2 \\
					3 & 4
					\end{pmatrix}
					$$
				</section>

				<section data-markdown>
					<script type="text/template">
						### Mathjax support with markdown

						$$
						\begin{pmatrix}
						1 & 2 \\
						3 & 4
						\end{pmatrix}
						$$

					</script>
				</section>

The second just displays the four numbers in a row.

pedritomelenas:
It seems that the problem with arrays is solved if we add three backslashes instead of two in markdown for new row.

				<section>
					<h3>Mathjax support without markdow </h3>

					$$
					\begin{pmatrix}
					1 & 2 \\
					3 & 4
					\end{pmatrix}
					$$
				</section>

				<section data-markdown>
					<script type="text/template">
						### Mathjax support with markdown

						$$
						\begin{pmatrix}
						1 & 2 \\\
						3 & 4
						\end{pmatrix}
						$$

					</script>
				</section>

This is probably a problem of markdown parsing.

Many thanks for such a nice work!

[...]

PR #1205, opened 2015-04-16
Allow tex parsing in <code> tags

gothma:
This adresses #702.

Before: `$\sum_{m} \sum_{d}$`  
screen shot 2015-04-16 at 13 38 26

After: ``` `$\sum_{m} \sum_{d}$` ```  
screen shot 2015-04-16 at 13 38 46

Just using $ as delimiter in markdown document fails since the markdown
parser unknown to the dollar syntax will try to interpret underscores.
Putting the $ delimented formula in backticks will cause the markdown
parser to put the tex-code with the $ delimiters into a code block.
The texcode will then be unchanged. This patch allows for mathJax to
interpret and automagically display all tex-formulas.

### DIFF
--- a/plugin/math/math.js
+++ b/plugin/math/math.js
@@ -14,7 +14,10 @@ var RevealMath = window.RevealMath || (function(){
 
 		MathJax.Hub.Config({
 			messageStyle: 'none',
-			tex2jax: { inlineMath: [['$','$'],['\\(','\\)']] },
+			tex2jax: {
+				inlineMath: [['$','$'],['\\(','\\)']] ,
+				skipTags: ['script','noscript','style','textarea','pre']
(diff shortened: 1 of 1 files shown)


hakimel:
Great. Thanks!

zoof:
Not sure why but this fix is not working for me, even with the example from https://github.com/hakimel/reveal.js/issues/702 (http://pastebin.com/zqSm4d4h) or the above example: `$\sum_{m} \sum_{d}$`.

gothma:
Have you enclosed the formula with backticks? (Notice the difference between before and after)

zoof:
Ah!  The backticks looked like a typo which I promptly ignored....

PR #1218, opened 2015-05-01
Update multiplexing doc for easier self hosted socket.io server and setup.

ArthurMaroulier:
Add a section with more details for self hosted socket.io server configuration and master / clients presentation configuration.


paambaati:
+1 for this. The documentation should clearly state this.

PR #1220, opened 2015-05-03
Fix last row header having a border-bottom

flying-sheep:
`<tbody/>` is automatically inserted by the browser, so we can guarantee that this affects all ths and tds in the last row of the table *body* and the table head is left untouched.


hakimel:
Thanks! This was merged into dev https://github.com/hakimel/reveal.js/commits/dev

PR #1228, opened 2015-05-13
add missing dependencies in package.json

lenciel:
I saw errors after creating a clean fork and ran `grunt serve` command. So I added these two lines in package.json.


hakimel:
Neither of those two should be required, all grunt tasks work for me and I use the default Gruntfile included in the project. 

Maybe you can give it another go? Pretty much all dependencies have updated since this.

PR #1238, opened 2015-05-24
Change keyboard bindings for all slides/multiple variables in reveal.js

ktmduke:
Hi, 
I’m using reveal.js for a presentation of some short video clips (1 clip per slide). I’d like to use the spacebar (key=32) to play/pause the videos and thus I tried to change/override the default keyboard bindings in reveal.js using the following code, which actually works fine for the first video (first slide) but not for the others. They all have the same ID (“video”). How can use the spacebar to play/pause all the videos in the presentation? Thanks for your help!
Reveal.configure({
  keyboard: {
    13: 'next', // go to the next slide when the ENTER key is pressed
    27: function() {}, // do something custom when ESC is pressed
    32: function(){ 
          if (video.paused == true) video.play();
          else video.pause();
          }
     }
});


hakimel:
This issue tracker is for feature request and bug reports. If you want to ask a general question or get help please try Stack Overflow.

In short you'll need to iterate over all videos and pause them. When resuming playback you can get the current slide with  `Reveal.getCurrentSlide()` and then find the current video within that.

PR #1257, opened 2015-06-05
Accessibility: controls as buttons, not divs

marcysutton:
Occasionally screen reader users get into a situation where the application's keyboard commands don't do what they expect, most notably in JAWS. To give users more control over the application, this PR converts the controls into button elements instead of divs so they are reachable and operable from a keyboard.

The changes are isolated to `reveal.js` and `reveal.scss`. @hakimel I can add test coverage if that makes sense, but it seemed like overkill for this.

### DIFF
--- a/css/reveal.css
+++ b/css/reveal.css
@@ -156,58 +156,61 @@ body {
   bottom: 10px;
   -webkit-user-select: none; }
 
-.reveal .controls div {
+.reveal .controls button {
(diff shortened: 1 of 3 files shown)


hakimel:
Thanks! This has been merged into the dev branch and will be released with the next version.

I updated the controls selectors in the themes as well so that they still work.

PR #1281, opened 2015-06-24
Update README to include socket.io 1.3.5 examples

### LINKED ISSUE #1276: Multiplex no longer works with example given in README due to socket.io version discrepancy with revealjs.jit.su
The README currently gives this example for a multiplex configuration:

```javascript
Reveal.initialize({
	// other options...

	multiplex: {
		// Example values. To generate your own, see the socket.io server instructions.
		secret: '13652805320794272084', // Obtained from the socket.io server. Gives this (the master) control of the presentation
		id: '1ea875674b17ca76', // Obtained from socket.io server
		url: 'revealjs.jit.su:80' // Location of socket.io server
	},

	// Don't forget to add the dependencies
	dependencies: [
		{ src: '//cdnjs.cloudflare.com/ajax/libs/socket.io/0.9.16/socket.io.min.js', async: true },
		{ src: 'plugin/multiplex/master.js', async: true },

		// and if you want speaker notes
		{ src: 'plugin/notes-server/client.js', async: true }

		// other dependencies...
	]
});
```

Evidently, the socket.io server at revealjs.jit.su has been upgraded to a 1.x version, with which the 0.9.16 client library is incompatible, resulting in an HTTP 400 (Bad Request) error.

Updating the client socket.io library to 1.3.5 makes the problem go away; thus the correct socket.io source URL appears to be `//cdn.socket.io/socket.io-1.3.5.js`.

fghaas:
Fixes #1276. Doc only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -817,7 +817,7 @@ Reveal.initialize({
 
 	// Don't forget to add the dependencies
 	dependencies: [
-		{ src: '//cdnjs.cloudflare.com/ajax/libs/socket.io/0.9.16/socket.io.min.js', async: true },
+		{ src: '//cdn.socket.io/socket.io-1.3.5.js', async: true },
 		{ src: 'plugin/multiplex/master.js', async: true },
 
 		// and if you want speaker notes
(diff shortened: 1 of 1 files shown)


hakimel:
Makes sense to update but will want to test the potential conflict with the notes plugin that you mentioned in https://github.com/hakimel/reveal.js/issues/1276#issuecomment-114813103. Might not be an issue at all but want to eliminate the risk.

fghaas:
@hakimel I agree. I, for one, wouldn't at all mind to be able to use both notes and multiplex together. :)

PR #1067, opened 2014-12-03
Add bower.json

teawithfruit:

PR #1282, opened 2015-06-24
Added bower.json, allowing to import reveal.js using bower

burnpanck:
With this little addition, you can build your reveal.js presentation using bower. I set `private=true`, in case it is not intended to actually push reveal.js to the bower repositiory. Even then, one can specify the github repo as the dependency in your presentation's `bower.json` as follows:

```json
"dependencies": {
    "reveal": "git://github.com/hackimel/reveal.js#dev"
}
```

Note that I also removed the `.js` from the name both in `bower.json` and `package.json`, to conform to the npm and bower guidelines.


hakimel:
Hey. This was started in #1067 a while back, I just got around to making some revisions to the file and have added it in master. Hope it works!

Issue #1261, opened 2015-06-11
Slide background overlap in the pdf export


saimn:
Following #1252, if I use an empty slide with a background image (to show a fullscreen image), the background image overlaps with the next slide content in the pdf export:
```
<section data-background="https://s3.amazonaws.com/hakim-static/reveal-js/image-placeholder.png">
</section>
<section data-background="https://s3.amazonaws.com/hakim-static/reveal-js/image-placeholder.png" data-background-repeat="repeat" data-background-size="100px">
....
```

![2015-06-11-103436_1680x1050_scrot](https://cloud.githubusercontent.com/assets/311639/8103052/aff0b402-1025-11e5-9198-9dfa175b3c5c.png)

This can be avoided by adding some fake content in the empty section (`<div>&nbsp;</div>`), but it can be surprising when discovering reveal.js

ddeboer:
I’m having this (or a similar) problem too. You can see that (black) background color from the previous slides slightly overflows to the next (grey) slide.

![screen shot 2015-06-28 at 21 14 47](https://cloud.githubusercontent.com/assets/89267/8397869/be7f0d58-1dda-11e5-83d7-9bdac3cdace1.png)

@saimn There seem to be no empty sections in the above slides. What exactly do you mean with adding the empty div?

obilodeau:
I was bitten by both of your PDF rendering bugs today. I looked at the code and tried to fix them but it seems that even though I triggered the `@media print` in the developer console I couldn't really emulate the print view and inspect how CSS affected the HTML tags and interactively debug it.

I could get a proper render with phantomjs and the [print-pdf phantomjs plugin](https://github.com/hakimel/reveal.js/blob/master/plugin/print-pdf/print-pdf.js) but links weren't working...

Long story short, I found about [decktape](https://github.com/astefanutti/decktape) which instruments phantomjs to crawl the slides and dump them into a PDF and it fixed all of my problems.

I'll open a PR to suggest trying it in the README under the print-pdf option.

ddeboer:
Thanks @obilodeau, using decktape worked for me!

opensource21:
decktape has a different focus. It creates a page for each fragment which is shown on a slide. More painful I wasn't able to create output in the format of DIN A4. So it would be nice if the issue could be fixed.

jdwolin:
Having this same issue of overlapping slides on PDF export even with content.  @obilodeau, can you explain more specifically how you got decktape to work in place of the default PDF export package?

[...]

PR #1287, opened 2015-06-29
README: suggest decktape as an alternative for PDF rendering

obilodeau:
It works around the current PDF rendering bugs like #1261. I have experienced other PDF rendering issues where the text's vertical alignment varies from slide to slide. I guess it's just simpler to instrument a browser and where the decktape project shines is that they use a fork of phantomjs to fix issues where links were not properly kept in PDF renders.

I'm open to rewording of the sentence. At first I stated why it was a superior alternative for pixel perfect PDF rendering of the slides compared to the print-pdf option but decided that less text was better.

I enjoy reveal.js so much. Thanks for your work!
p.s.: for the record, I am not affiliated with decktape

### DIFF
--- a/README.md
+++ b/README.md
@@ -704,6 +704,8 @@ Here's an example of an exported presentation that's been uploaded to SlideShare
 
 ![Chrome Print Settings](https://s3.amazonaws.com/hakim-static/reveal-js/pdf-print-settings.png)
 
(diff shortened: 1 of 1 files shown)


hadim:
I didn't know and it fixes all the issue I have with the current pdf rendering script.

Awesome ! Thank you !

hakimel:
Haven't tested decktape myself yet but looks great. Thanks for the PR!

PR #1295, opened 2015-07-02
Gruntfile.js: watch tasks restructured

jzgdev:
Issues with livereload on OSX / Chrome.
Restructured gruntfile according to `grunt-contrib-watch` documentation


hakimel:
Hm, moving the definition further down inside of the same object definition shouldn't make a difference. Was there something in their documentation specifically about the order?

PR #1310, opened 2015-07-21
solves early access error with video element

teawithfruit:
This will maybe solve the "InvalidStateError: An attempt was made to use an object that is not, or is no longer, usable" error in firefox.


hakimel:
Good stuff, thanks!

Issue #1319, opened 2015-08-03
Sections with id's not in all lower case break history functionality


gruber76:
If a section has an id with one or more capital letters, the user cannot use the back button to return to that slide.

gruber76:
I'm not sure if the "toLowerCase" call in `writeURL` was necessary for some functionality, but taking it out appears to fix this issue (and has the side effect of prettier URLs.)

hakimel:
Fixed by @gruber76 in #1320

Issue #836, opened 2014-03-17
Safari switching slides broken if id of a slide contains cyrillic


certainlyakey:
In Safari, both mobile and desktop, if you have some cyrillic symbols inside a slide's id, strange characters appear in address bar instead of proper id value and it's possible to switch only a few slides - after that the slideshow jumps to the last slide.

I noticed that because I use Pandoc to create presentations and it automatically places whatever is the heading of the slide into the id attribute of a slide section tag.

Here's a [test presentation](http://x.co/3rS2A).

As a workaround, it would be great to have an option to turn off the section ids as hashes and leave only slides' numbers.

PR #1320, opened 2015-08-03
Update reveal.js

gruber76:
To address issue #1319, removed toLowerCase call which was added when fixing issue #836.  (Disallowing lowercase URLs does not appear to have been related to the original issue.)


hakimel:
Thanks for taking the time to look into this. I agree it doesn't seem like there's any reason for the string to be forced to lowercase. I've merged your commit into the dev branch: https://github.com/hakimel/reveal.js/commits/dev

In the future, please submit pull requests towards dev rather than master. Thanks!

PR #1325, opened 2015-08-08
resolves #1324 allow display property to be configured

### LINKED ISSUE #1324: Allow display property of visible slide to be configured
The flexbox model is a very useful way to layout slide content. However, at the moment, it's not possible to use the flexbox model for slides because reveal.js is explicitly setting the display CSS property on the current slide to "block". (See https://github.com/hakimel/reveal.js/blob/master/js/reveal.js#L2724).

Please allow the display property of the current slide to be configured so that other slide layouts can be used. In particular, allow the value to be "flex".

I propose the name `display` for this setting.

``` javascript
Reveal.initialize({
  display: 'flex',
  // ...
});
```

An alternative is to detect the original display value using `getComputedStyle` and reuse that value when setting the display CSS property to reveal the current slide.

mojavelinux:
- allow display property apply to current slide to be configured
- ~~set flex-direction to column by default in case flexbox is used~~

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -45,6 +45,9 @@
 			minScale: 0.2,
 			maxScale: 1.5,
 
+            // Value of the display CSS property applied to current slide to make it visible
+            display: 'block',
+
 			// Display controls in the bottom right corner
 			controls: true,
 
@@ -2721,7 +2724,7 @@
 	function showSlide( slide ) {
 
 		// Show the slide element
-		slide.style.display = 'block';
+		slide.style.display = config.display;
 
 		// Media elements with data-src attributes
 		toArray( slide.querySelectorAll( 'img[data-src], video[data-src], audio[data-src]' ) ).forEach( function( element ) {


mojavelinux:
I can confirm that setting `display: 'flex'` in the configuration does not affect the appearance of the sample slide deck, either in the HTML or the print view. The benefit, of course, is that flexbox can now be used to position elements on the screen.

mojavelinux:
Actually, there is a small difference. The center alignment no longer works. It's necessary to add `align-items: center` to the CSS to match the original behavior. But, of course, that's really up to the theme author. For that reason, I'm going to back out the change to adding `flex-direction` to `reveal.css`.

hakimel:
Makes sense to allow this to be configured. Your commit has been merged into the dev branch, and I added the new config value to the docs as well.

mojavelinux:
:+1:

pstjohn:
One issue with this approach is that its hard to have vendor prefixes handled appropriately. Specifically for safari, you'd really want.

```css
   display: -webkit-flex;
   display: flex;
```

PR #1328, opened 2015-08-14
Added formerly undocumented "cube" transition to demo

dainiak:


hakimel:
Thanks but the cube and page transitions are no longer mentioned there since they are deprecated.

Issue #1133, opened 2015-02-15
Highlight specified line(s) in source code


szpak:
In some cases it is very useful to be able to highlight specified line(s) in source code by giving them other background (e.g. yellow).

The mentioned effect achieved with remark.js: http://jsfiddle.net/szpak/059x3ygs/1/

Delapouite:
The highlighting is handled by https://highlightjs.org so I think this request should be open on their repo.
I would also find this useful, especially if you could highlight only a portion of a line, for example just a keyword.

szpak:
Good idea. Reported as [#740](https://github.com/isagalaev/highlight.js/issues/740).

wsmoak:
Did anyone get this working?  https://github.com/isagalaev/highlight.js/issues/740 is closed with the advice that html is not escaped inside a `<code>` tag.

But when I try this as described on http://jsfiddle.net/tovic/059x3ygs/2/
```
<section>
            <pre><code data-trim>
{application, elixir,
[{description, "elixir"},
<mark class="highlight-block">{vsn, "1.1.0-20150912"}</mark>
 {modules, [
        elixir
  ]},
      </code></pre>
    </section>
```

It briefly flashes in yellow, but then I get the `<mark>` tag printed literally into the slide, not the line highlighted like I wanted.

wsmoak:
Based on https://github.com/hakimel/reveal.js/blob/master/plugin/highlight/highlight.js#L14 it looks like you should be able to add a `data-noescape` attribute to avoid the html inside the `code` tag getting escaped.

Unfortunately, it's not working for me.  I changed the line above to 
```<mark data-noescape class="highlight-inline">{vsn, "1.1.0-20150912"}</mark>```
and it still flashed and then displayed the text rather than staying highlighted.

If I comment out lines 15-17 of reveal.js's `plugin/highlight/highlight.js` THEN it works as expected and highlights the indicated line.

wsmoak:
I opened https://github.com/hakimel/reveal.js/issues/1352 to add `data-noescape` and `<mark>` to the README.  

I also noticed this plugin listed on the wiki that might be useful (I haven't tried it):
Focus on specific lines of code blocks - https://github.com/d10/reveal-code-focus

demoneaux:
Closing this for now as #1352 has already been merged, and there’s already a plugin for that (maintained by me!).

(The repo for my plugin is now at https://github.com/demoneaux/reveal-code-focus/.)

PR #1352, opened 2015-09-13
Add the data-noescape attribute and example

wsmoak:
Documents the data-noescape attribute and shows an example of `<mark>`ing a line of code with highlight.js

This should solve https://github.com/hakimel/reveal.js/issues/1133

(The guidelines for contributing say to target the dev branch, but master seems to be ahead of it.  Let me know if I need to re-do this.)


hakimel:
Thanks – that should definitely be included in the documentation.

PR #1362, opened 2015-09-20
Update the theme README.md

GuGuss:
The `grunt css-themes` command now loops through all Sass files. There is no need to add the newly created file to the task list.


hakimel:
Thanks, totally forgot to update that.

Issue #1333, opened 2015-08-21
Text, other matter cut off when printing in Chrome


lordsutch:
The right margin seems to be set in such a way that text overruns the margin and is cut off. This can be illustrated on the example site, http://lab.hakim.se/reveal-js/, on the second page.

Example output (note the "w" on the word "show"):
![cutoff](https://cloud.githubusercontent.com/assets/2024917/9417226/77826e00-4818-11e5-9444-fa7baf1f2162.png)

This also seems to affect images floated right.

It appears the problem is line 144 of css/print/print.css: replacing "100%" with "auto" seems to fix the problem.

hakimel:
Fixed in #1381

PR #1381, opened 2015-10-03
Fix text being cut off in Chrome when printing.

lordsutch:
This pull request fixes issue #1333 .


hakimel:
Looks much better. Thanks!

PR #1382, opened 2015-10-03
Add autoplay support for asciinema

fghaas:
[Asciinema](https://asciinema.org/) is a popular service allowing users to record terminal screencasts ("asciicasts") and play them back in a react.js based player, enabling easy after-the-fact manipulation of color themes, font size, playback speed etc.

Since asciinema/asciinema.org@2913c045, Asciinema supports an `asciicast:play` event that enables autoplay. Fix up reveal.js in the same way as previously for Vimeo and YouTube.

This patch should also apply cleanly to master.

For background, see also asciinema/asciinema.org#175.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -2996,6 +2996,10 @@
 		else if( /player\.vimeo\.com\//.test( iframe.getAttribute( 'src' ) ) && iframe.hasAttribute( 'data-autoplay' ) ) {
 			iframe.contentWindow.postMessage( '{"method":"play"}', '*' );
 		}
+		// asciinema postMessage API
+		else if( /asciinema\.org\/api\/asciicasts\//.test( iframe.getAttribute( 'src' ) ) && iframe.hasAttribute( 'data-autoplay' ) ) {
+			iframe.contentWindow.postMessage( ['asciicast:play'], '*' );
+		}
 		// Generic postMessage API
 		else {
 			iframe.contentWindow.postMessage( 'slide:start', '*' );


Delapouite:
Hi

Youtube and vimeo support also appears in the stopEmbeddedContent function just below your contribution,
Do you think asciinema should be added there as well?

fghaas:
@Delapouite I'm afraid that wouldn't make sense. Asciinema has an API method that allows playing an asciicast, but not one that allows stopping or pausing it. See https://github.com/asciinema/asciinema.org/blob/master/app/views/api/asciicasts/show.html.slim if you want to take a look.

demoneaux:
This seems quite specific, and is better suited to a plugin IMO. YouTube and Vimeo are used by a large number of people, while I think much less people use Asciicinema. Perhaps this is better suited to a plugin.

fghaas:
A completely unintrusive three-line patch, and you want a plugin for that? Are you sure?

demoneaux:
IMO functionality which isn't used widely shouldn't have a need to be in the core scripts. Although this might be a 3 line patch, if everyone were to submit small patches like this for every single service, the file would become extremely bloated. In the end though, it's up to @hakimel to make the decison.

fghaas:
Look, I don't mind, I've had this in my own fork for a while anyway and don't mind keeping it that way. Also, seeing as this patch has been sitting here for 5 months I hadn't expected it to go in. Let's close this.

PR #1385, opened 2015-10-06
fix(plugin:highlightjs): update to version 8.8

thib3113:
script tag in html code crash with 8.6, it's fix now


hakimel:
Thanks for the PR. I was about to merge this but noticed there was a newer release (8.9.1) so I went ahead and updated to that instead.

PR #1391, opened 2015-10-08
Fixed #1379 - config keyboard now disables 's' key

### LINKED ISSUE #1379: Setting keyboard: false does not disable the s key
I am using the keyboard for another JS library but RevealJS fails to disable the s key when the keyboard: false setting is used. Any reason why?

rohithpr:


hakimel:
Good one, thanks!

Issue #1390, opened 2015-10-08
README does not provide full list of transitions


adunning:
There are references to page and cube transitions later in the document, but the only transition styles listed as configuration options are `none/fade/slide/convex/concave/zoom`.

hakimel:
The page and cube transitions used to be in the readme but have since been deprecated due to how inflexible they were. They're still in the CSS for backwards compatibility though.

adunning:
Thanks; I hadn't realized this. There is still one reference to the removed transitions: is this intentional?

hakimel:
That was unintentional – just removed it in 35462a424c4899747aa2079101372c66ac14d211.

adunning:
Great; thank you!

PR #1398, opened 2015-10-13
Fix Issue #1390 by adding full list of transitions

billglover:
Add the full list of transitions to documentation and comments.

The full list of transitions was taken from reveal.css.


hakimel:
Thanks for the PR. The "linear" and "default" transitions were renamed to "slide" and "convex". The old names remain in the CSS for backwards compatibility but should not be referenced in the readme.

PR #1401, opened 2015-10-14
Move up Presentation Size section.

adunning:
This brings together all configuration options.


hakimel:
Good call, thanks!

adunning:
Thank *you* for reveal.js!

Issue #1022, opened 2014-09-30
class=stretch doesn't work in print-pdf


VladimirAlexiev:
[Stretching](https://github.com/hakimel/reveal.js/#stretching-elements) is a very nice feature: I very often have an intro para, followed by a large image that I want to take the rest of the page.

It works for normal mode, but unfortunately not for print-pdf.

Inspecting with Chrome I see:
- screen:
  `<img src="./img/GVP-architecture.png" alt="GVP-architecture.png" class="stretch" style="width: 977.91646778043px; height: 733px;">`
- print-pdf: 
  `<img src="./img/GVP-architecture.png" alt="GVP-architecture.png" class="stretch">`
  and the image is too small

I work around this by putting for every wide image:
`class="stretch" style="width:1000px"`
(I run at width=1200, height:925). It stretches optimally for screen and still works ok for print.

For narrower images, I have to specify a smaller number to avoid page overflow

obama:
+1

cjoneill:
Rather than add markup to each image, I'm currently hacking around this as follows:

```
<style type="text/css">
  .print-pdf .stretch {
    max-width:100%;
  }
</style>
```

ramezrafla:
This issue is still open. Upon inspection of the code, it seems layout of the slides is done real-time when the slide is shown in browser: ```` slide(..) --> layout() --> layoutSlideContents(..) ````

There is therefore no easy solution to this problem, except writing code that traverses all slides when in PDF mode and runs layoutSlideContents(). 

Given the original author has an online service to create those slides, it is likely that the size information is added at creation time, vs. manual slides creation which expects images to automatically size.

This issue will therefore likely need to be solved by the community. We will look into it too at our end. Will require deeper understanding of the code (or just write new functions for PDF only)

ramezrafla:
I believe I solved it (or at least am almost there, works at my end). 

In reveal.js, add this at the end of function ````setupPDF```

````		layoutSlideContents(pageWidth*0.9,pageHeight*0.9,10); ````

The numbers may need to be tweaked a bit, but looks great at my end

hakimel:
Fixed in dev in #1844

PR #1416, opened 2015-10-30
Support PDF printing for img.stretch

ramezrafla:
Hakim, this is a minor change to fix the non-resizing of images with class 'stretch' when printing PDF. The numbers may need to be tweaked but looks good at my end. This refers resolves issue #1022

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -555,6 +555,9 @@
 			}
 		} );
 
+		// to resize images with class stretch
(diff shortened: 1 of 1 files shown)


riezebosch:
This works for me, please merge!

edmcdonagh:
This works for me too, and is very useful with large images that are constrained by 'stretch' in the standard view.

hakimel:
Merged a fix for this in #1844, it's mostly the same as the change in this PR but using the explicit page width/height instead of a scaled down page width/height.

PR #1419, opened 2015-11-02
fixed loading

teawithfruit:
fixed marked.js loading with require.js. the markdown plugin rises an error, if marked.js is not loaded before using it.


hakimel:
Your commit has been merged into the dev branch and will go out with the next release, thanks!

PR #1421, opened 2015-11-03
README - remove script tag from markdown section

petrblaho:
\<script type="text/template"\> is not needed when using \<section data-markdown\> as
shown in index.html.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,17 +37,15 @@ Markup hierarchy needs to be ``<div class="reveal"> <div class="slides"> <sectio
 
 ### Markdown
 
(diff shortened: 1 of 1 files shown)


hakimel:
The template wrapper is needed so that the Markdown content isn't parsed by the browser before we can pick it up using JavaScript. For example, neither of the following would work without the wrapper:

```html
<section data-markdown>
	<script type="text/template">
		<http://slides.com>

		```<b>Hello world</b>```

		<u>test</u>
	</script>
</section>
```

Issue #1418, opened 2015-11-01
Synchronous XMLHttpRequest on the main thread is deprecated AND makes some CDN usage difficult


ivanoats:
I get the warning Synchronous XMLHttpRequest on the main thread is deprecated. 

The full warning is:
`d2q4nobwyhnvov.cloudfront.net/dffb9bbc-edbd-4e7d-a0c1-546a50ade31e/tI-NcHSnnRPTlg/plugin/markdown/markdown.js:254 Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user's experience. For more help, check http://xhr.spec.whatwg.org/.`

I guess this is because I am using Amazon's CloudFront CDN.

It's in line 254 of plugin/markdown/markdown.js. 

What is your preferred way of changing this sync call to async?  Interested in a PR? I could get one started with some guidance.

PR #1428, opened 2015-11-10
Changed XMLHTTPRequest mode to asynchronous.

sloan-848:
Addresses issue #1418 by changing mode to asynchronous for XMLHTTPRequest call.


s-bernard:
This does not work as convertSlides is called before processSlides is actually finished.

s-bernard:
I created #1458 which fixes the problem.

PR #1431, opened 2015-11-11
The external markdown file can not be rendered when including highlight.js directly.

snowyu:
see the tests/test-markdown-external.html for detail.


test/test-markdown-external.js
```diff
@@ -0,0 +1,19 @@
+
+
+Reveal.addEventListener( 'ready', function() {
+
+	QUnit.module( 'Markdown' );
+
+	test( 'Vertical separator', function() {
+		strictEqual( document.querySelectorAll( '.reveal .slides>section>section' ).length, 2, 'found two slides' );
```
hakimel: Looks like this is giving a false positive. The md file parses into a presentation where "Slide 1.1", the code sample and "Slide 1.2" are all on the same slide. The selector still returns a length of 2 since "Slide 2", which is intended to be a horizontal slide, actually becomes a vertical slide.

snowyu: I just repair the wrong arguments of the highlighter's invocation. No touch
other codes.

hakimel: Got it. Looking at it again I can see the vertical separator test was carried over from the `test-markdown.js` file, I'll revise it to match the simple.md that `test-markdown-external.js` tests against.

hakimel:
All set – your commits have been merged into the dev branch. Please submit pull requests directly towards that branch next time, thanks! https://github.com/hakimel/reveal.js/commits/dev

Revised the test as discussed above e4634f20e8be5075882b3cc49a169e0fa1ae8f6a.

PR #1434, opened 2015-11-13
stops parallax failing if only the single horizontal slide and multiple vertical

mikemellor11:
Check for 0 division, if so set to horizontalOffsetMultiplier to 0, this allows for presentation's that are all vertical slides


hakimel:
Thanks!

PR #1436, opened 2015-11-16
Fix misleading quotes for parallax config.

adunning:
As reveal.js expects an integer rather than a string, quotes should not be used.


hakimel:
Good catch. They shouldn't be blank though, I changed them to `null` since that's what reveal.js defaults to. 833622edcd45a7721f7a76d93cc12a1284f60a0d

adunning:
Great; thank you!

PR #1437, opened 2015-11-18
Add an option to auto-slide right only

der-michik:
I needed to prepare a presentation that runs automatically in an infinite loop with a few "bonus slides" that further explain certain topics within the presentation. However, auto-sliding per default always navigates to the next slide which may be below the current one.

With the autoSlideRight option enabled, auto-sliding will ignore the basement levels and always navigate to the right. However, the user can always intervene and seamlessly bring up the additional slides with the bonus content.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,11 @@ Reveal.initialize({
 	// Stop auto-sliding after user input
 	autoSlideStoppable: true,
 
+	// When auto-sliding is active, do always proceed to the right
+	// instead of the next slide which may be below (useful for
+	// infinite loop presentations with hidden "bonus slides")
+	autoSlideRight: false,
+
 	// Enable slide navigation via mouse wheel
 	mouseWheel: false,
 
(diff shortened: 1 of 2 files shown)


hakimel:
Thanks for the PR. This is an interesting idea, though I think it could be made a bit more flexible. What if we made the config option `autoSlideMethod` instead so you can switch to any navigation method?

```
// Only step to the right
Reveal.initialize({ autoSlideMethod: Reveal.navigateRight })

// Go backwards
Reveal.initialize({ autoSlideMethod: Reveal.navigatePrev })

// Go backwards
Reveal.initialize({ autoSlideMethod: function() {
  // custom behavior
} })
```

Would be pretty straight forward to add, we just need to replace navigateNext here and make sure it defaults to using `navigateNext`: https://github.com/hakimel/reveal.js/blob/master/js/reveal.js#L3691

der-michik:
Maybe like this?

hakimel:
This is great – thanks!

PR #1445, opened 2015-11-22
Fragments

kewitz:
Hakim, I added four new fragments transitions effects that are basically about fading into a direction.


hakimel:
Nice! Those look great.

PR #1450, opened 2015-11-30
Chromium users need to check Background graphics option when printing.

Quintasan:
I have tried printing my slides as PDF today and I was wondering why the hell all of my slides are white. Turns out you need to check **Background graphics** option in Chromium when printing.


hakimel:
Thanks, I'm sure that'll be helpful to others.

PR #1451, opened 2015-12-01
added version to Reveal

SethosII:
version should be able to be queried by plugin


hakimel:
That's a good idea. However the variable you added was in local scope and couldn't be queried by plugins from the outside. I went ahead and revised it so the version is now accessible through `Reveal.VERSION`.

PR #1456, opened 2015-12-04
Shrink/Grow fragment blinking on span

AlexisBRENON:
When applied to `span` elements, the shrink/grow fragment style is applied but the style is restored to the normal state immediately. Changing the `display` property to `inline-block` seems to fix this issue.


AlexisBRENON:
Yes of course, I don't know why the second and third commits are here. I'll fix this ASAP.

AlexisBRENON:
Fixed

PR #1458, opened 2015-12-07
Set asynchronous load of external markdown

s-bernard:
Set xhr.open as asynchronous while loading an external markdown file.
This fixes https://github.com/astefanutti/decktape/issues/12 which then resolves all pdf-export problems.

### DIFF
--- a/plugin/markdown/markdown.js
+++ b/plugin/markdown/markdown.js
@@ -205,7 +205,7 @@
 	 * multi-slide markdown into separate sections and
 	 * handles loading of external markdown.
 	 */
-	function processSlides() {
+	function processSlides(callback) {
 
 		var sections = document.querySelectorAll( '[data-markdown]'),
 			section;
@@ -251,7 +251,8 @@
 					}
 				};
 
-				xhr.open( 'GET', url, false );
+				xhr.onload = callback
+				xhr.open( 'GET', url, true );
 
 				try {
 					xhr.send();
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks for the PR. I gave this a try but wasn't able to get it to work. When I try opening the Markdown example presentation (`/plugin/markdown/example.html`) the first slide – which should be loading external Markdown – is empty.

We'll also need to remove the `Reveal.slide(0)` call at the end since that currently overrides all deeplinks.

astefanutti:
It turns out the original has been fixed in PhantomJS with ariya/phantomjs#13908.

hakimel:
If I'm understanding this correctly, the issue has been fixed at the source so there's no longer any need for this change. Let me know if I'm wrong and I'll reopen.

fiddur:
It is a problem in firefox at least, just like in #1418 the xhr needs to be async.

tebanep:
Hello @hakimel, I'm still experiencing this error both on chrome 58 and in firefox 53:

`
Synchronous XMLHttpRequest on the main thread is deprecated because of its detrimental effects to the end user’s experience. For more help http://xhr.spec.whatwg.org/
`

On plugins/markdown/markdown.js line 246:

`
xhr.open( 'GET', url, false );
`

Should be asynchronous.

mosabua:
This is definitely still a problem. E.g. we have a slide deck at https://concord.walmartlabs.com/training/concord-1-beginner.html#/ which works fine on all browsers apart from Firefox .. and there if fails because of the synchronous load of the markdown js file.

PR #1465, opened 2015-12-22
Add "background div" to allow HTML backgrounds

upayavira:
This simple patch allows backgrounds to be specified as HTML. This has many benefits - you can include diagrams as backgrounds - better still, you can add headings as backgrounds which will appear correctly at the top of the page. To use it, you create a hidden div outside the section and reference it with a data-background-div tag via its ID:

    <section data-background-div="my-heading">
       <p>My content</p>
    </section>
    <div id="my-heading" style="visibility: hidden">
       <h1>My heading</h1>
    </div>

When the background is generated, the div is cloned and added to the background div that Reveal already creates. "if" is used rather than "else if" as this allows this feature to be used alongside data-background-image, which can create some nice effects.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -2870,7 +2870,8 @@
 				var backgroundImage = slide.getAttribute( 'data-background-image' ),
 					backgroundVideo = slide.getAttribute( 'data-background-video' ),
 					backgroundVideoLoop = slide.hasAttribute( 'data-background-video-loop' ),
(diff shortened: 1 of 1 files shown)


Delapouite:
Nice! What about using a proper `<template>` tag instead. Could it work in your case?

upayavira:
Yes, it seems a <template> tag would work also - effectively it is a div with visibility set to hidden by default, which is all I am proposing above.

PR #1466, opened 2015-12-27
Add speaker notes to keyboardShortcuts

ismith:
This improves discoverability; I knew reveal.js had this, but it took some
digging to figure out how to use it.


hakimel:
Good idea, however the speaker notes plugin is optional so the shortcut should only be included when the plugin is present on the page. I implemented it to work that way in 06cdd9b

PR #1472, opened 2016-01-06
Add the grunt-cli as a dependency instead of assuming global grunt.

vicapow:
Otherwise, `npm run start` doesn't work without first globally installing grunt.


hakimel:
Cool, thanks!

PR #1495, opened 2016-01-29
add custom renderer to markdown plugin

Neths:
Allow customization for markdown plugin renderer


hakimel:
It's now possible to configure marked via the new `markdown` config option, for example:

```javascript
Reveal.initialize({
  markdown: {
    smartypants: true
  }
})
```

PR #1502, opened 2016-02-10
Fix IE 11 crash when changing slides

nikolaik:
Assignment of whitespace to textContent triggers it, so always do trim()

IE bug report: https://connect.microsoft.com/IE/feedback/details/1483791/ie-crashes-when-viewing-any-page-on-tpnn-com

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -2216,7 +2216,7 @@
 		}
 
 		// Announce the current slide contents, for screen readers
-		dom.statusDiv.textContent = currentSlide.textContent;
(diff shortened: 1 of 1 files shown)


demoneaux:
This doesn't seem to be reproducible for me. Can you provide a reduced test case which fails for you?

nikolaik:
Forgot about this, sorry. I tried creating a minimal testcase, but could not reproduce the crash. Doing trim() before assigning to textContent solved the bug using production data.

hakimel:
Closing since the original issue can't be reproduced.

Issue #1089, opened 2014-12-28
shuffle slides


benjamingr:
Hey,

Are there any plans to support shuffling slide order or randomizing slide order?

If there are not - would you entertain a PR?

hakimel:
Thanks for the suggestion but doesn't feel like this is a common enough use case to merit adding to the framework. A small script ahead of reveal.js initializing should be able to do this.

demoneaux:
Closed in #1506.

benjamingr:
Thanks!

PR #1506, opened 2016-02-14
Introduce random config

lyndseypadget:
This is a config-based, built-in alternative to proposals like [this one](http://stackoverflow.com/questions/24809752/randomize-slides-in-reveal-js).  This PR addresses #1089, which is closed but perhaps worth reconsidering.

To enable:

				// Simply turning random on will use the entire range of top-level (horizontal) slides
				// Reveal.configure({ random: true });
				
				// Or you can specify an inclusive range of slide numbers
				Reveal.configure({
					random: {
						rangeStart: 5,
						rangeEnd: 8
					}
				});

				//Autoslide can be used with random
				//Reveal.configure({ autoSlide: 3000 });

Notes:

- Of course, random is off by default
- Can be generally enabled (true) or further specify a range of slides (rangeStart and rangeEnd)
- Slide range is inclusive
- Ranges can be manually overridden by any code calling navigateRandom
- Range will autocorrect if not in the proper order ([5,1] becomes [1,5])
- Random navigation will ensure you are not routed to the slide you're already on
- Can be used with autoslide on or off
- Randomized slides are always at the top (horizontal) level - never "basement" levels
- Fragments are still honored
- Transitions are only executed when navigating in a forward (right) direction

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -147,7 +147,10 @@
 			viewDistance: 3,
 
 			// Script dependencies to load
-			dependencies: []
+			dependencies: [],
+
+			// Randomize slides
+			random: false
 
 		},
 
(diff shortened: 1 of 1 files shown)


demoneaux:
Marking this as a duplicate of #1089 for now. @hakimel do you have any opinions about this?

hakimel:
Thanks for the PR and taking the time to implement this. I said in #1089 that this didn't seem like a common enough use case, but judging by the number of votes on Stack Overflow I was wrong.

I'd be happy to add support for random order but I don't think this is the right way to do it. The slides should be moved into random order in the DOM instead, for two reasons:
- It'll be a lot easier to implement (just a few lines of JS to run at startup)
- it's a bit strange to have the slide numbers and progress bar jump around as they do when the DOM order remains unchanged

hakimel:
Added what I mentioned above in 2625a9292eae1c13c609eabd6d71eef825b55c92

PR #1512, opened 2016-02-18
Added ability to show controls on background videos.

FuzzyWuzzie:
I like to use background videos with no text as fullscreen video slides. Sometimes however, I want controls for such a video.

This commit add the `data-background-video-controls` data attribute, which if given in something like this:

```html
<section data-background-video="vid/gap_of_disappointment.webm" data-background-color="#000000" data-background-video-controls></section>
```

will show native controls on the video.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -2889,6 +2889,7 @@
 					backgroundVideo = slide.getAttribute( 'data-background-video' ),
 					backgroundVideoLoop = slide.hasAttribute( 'data-background-video-loop' ),
 					backgroundVideoMuted = slide.hasAttribute( 'data-background-video-muted' ),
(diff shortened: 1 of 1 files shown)


demoneaux:
I’m not too sure about this. I think the idea of having background videos were to have non intrusive videos playing at the back while having some content on the slides. I think in your case just adding a video into the slide is fine. @hakimel opinions?

hakimel:
Thanks for the PR! Backgrounds are intended to be non-interactive since they can't always be reached due to overlap with slide content or presentation control elements. For example, the reveal.js progress bar is at a higher depth than the backgrounds so when I tried clicking the video controls it was captured by the progress bar and ended up switching slides.

I agree with @demoneaux that it's better to place a video directly in the slide in this case.

PR #1517, opened 2016-02-25
Added before and after event

teawithfruit:
To handle the content before and after sliding, I've added two events which can be used with the following code.

Reveal.addEventListener( 'beforeslide', function( event ) {
  console.log(event);
});

Reveal.addEventListener( 'afterslide', function( event ) {
  console.log(event);
});

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -3761,6 +3761,7 @@
 	}
 
 	function navigateLeft() {
+		beforeSlide();
 
 		// Reverse for RTL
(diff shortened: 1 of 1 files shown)


hakimel:
This is a useful addition. 

There shouldn't be any need to dispatch these events from the individual "navigate" methods though, right? Those methods all end up calling `slide()`, and you're already dispatching the same events from within the `slide()` function.

PR #1518, opened 2016-03-01
Add possibility to print speaker notes

wasabideveloper:
Print speaker notes on a separate page after each slide if `print-pdf` and `print-notes` are given in url.
reveal-print-notes


hakimel:
Hey! Thanks for the PR but we already support printing notes to PDF. This is made possible via the `shareNotes` config option which was added in reveal.js 3.2.0. With `showNotes` set to true, the notes will be visible on top of the presentation and be included in exports.

You can test this on the demo page, slide 10 has some notes: http://lab.hakim.se/reveal-js/demo.html?print-pdf&showNotes=true

wasabideveloper:
I know. But `shareNotes` is not suitable for much lines of notes. So it's a nice possibility to print the notes to a separate page. Often people like to print two versions of their presentation: One with notes and one without. So it's much more comfortable using a url parameter to print the notes instead of changing the config every time.

hakimel:
You're right that the current notes design doesn't allow for very long notes. We should try to improve that in the existing solution rather than adding a secondary way to print. Perhaps an additional config value that pushes printed notes to a separate page? Alternatively we could move notes to the right of the slide as a vertical column that spans the full height of the page and about 20-30% of the width.

All config parameters can be set as url params, so you can just add ?showNotes=true to the URL. This works with other properties too, four example you can turn on auto-sliding every five seconds by adding `?autoSlide=5000`.

wasabideveloper:
I didin't know about the possibility of setting config params by the url.
Under this circumstances I propose to generalize the `printNotes` feature and add a `notesLayout` option.
```javascript
notesLayout: 'none' // no notes shown
notesLayout: 'default' // = 'none'
notesLayout: 'inline' // like the current 'showNotes' behaviour
notesLayout: 'side' // Slide size resized to 70% and a notes bar of 30% width on the right
notesLayout: 'separatePage' // On the following page
```
This `layoutNotes` option simply should include a css file from `css/print/notes/`so the user can create it's own. E.g:
```javascript
notesLayout: 'notesWithCommentLines' // will include css/print/notes/notesWithCommentLines.css
```

egrosclaude:
I'm a teacher and I'm VERY much interested in the alternative "notes" you
mention. In fact, I've Q&D hacked the 'notes' plugin to turn reveal into
some sort of a courseware app with handsome results. But alas, I'm no js
programmer, so my solution is just an ugly hack. If I had the said plugin,
well written by an expert, and then, the ?showLecture-or-something
parameter, this would be... just Heaven.

On Mon, May 23, 2016 at 7:15 AM, Hakim El Hattab <notifications@github.com>
wrote:

> You're right that the current notes design doesn't allow for very long
> notes. We should try to improve that in the existing solution rather than
> adding a secondary way to print. Perhaps an additional config value that
> pushes printed notes to a separate page? Alternatively we could move notes
> to the right of the slide as a vertical column that spans the full height
> of the page and about 20-30% of the width.
>
> All config parameters can be set as url params, so you can just add
> ?showNotes=true to the URL. This works with other properties too, four
> example you can turn on auto-sliding every five seconds by adding
> ?autoSlide=5000.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/hakimel/reveal.js/pull/1518#issuecomment-220941226>
>



-- 
Eduardo Grosclaude
Universidad Nacional del Comahue
Neuquen, Argentina

hakimel:
I like the idea of a layout option for notes. I went ahead and implemented at least the start of it in 3111d3b1ae12af2580cb45a18da208146701a6fd.

Instead of a new config value, I'm re-using the existing showNotes option. Here's a list of what values you can set showNotes to:

| Value       | Description |
| :------------ | :-------------- |
| nil | Notes aren't shown. |
| false | Notes aren't shown. |
| true | Notes are shown with default layout ("inline"). |
| string | Notes are shown with the given layout. The layout is added as  `data-layout="yourLayoutName"` on the `.speaker-notes` element so that you can easily attach any custom styles. |

For now, the only supported layout is `separate-page` which will make notes appear on separate pages in PDF exports.

wasabideveloper:
:+1: That's great!

PR #1522, opened 2016-03-06
Fix query parameter issue in notes plugin.

### LINKED ISSUE #1392: Speaker notes won't show slides if presentation url contains question mark
When I have a reveal.js presentation opened with a ? in the URL, I can not get the slides in the speaker notes loaded, because the iframes url's params are appended with a ? and browsers do not like two question marks in an URL.

Suggested fix in /plugin/notes/notes.html.

change :
```javascript
var currentURL = data.url + '?' + params + '&postMessageEvents=true' + hash;
var upcomingURL = data.url + '?' + params + '&controls=false' + hash;
```
to :
```javascript
var urlSep = (data.url.indexOf('?') > 0) ? '&' : '?';
var currentURL = data.url + urlSep + params + '&postMessageEvents=true' + hash;
var upcomingURL = data.url + urlSep + params + '&controls=false' + hash;
```

demoneaux:
Closes #1392.

### DIFF
--- a/plugin/notes/notes.html
+++ b/plugin/notes/notes.html
@@ -288,9 +288,10 @@
 						'backgroundTransition=none'
 					].join( '&' );
 
+					var urlSeparator = /?/.test(data.url) ? '&' : '?';
 					var hash = '#/' + data.state.indexh + '/' + data.state.indexv;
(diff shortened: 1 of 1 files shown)


hakimel:
Looks good, thanks!

PR #1525, opened 2016-03-09
Added a description of the undocumented timer-reset feature

Timothep:
Hi, since the "click to reset" link only appear "on mouse over" in the speaker view, I only found it by accident after I tried to implement it myself. I somehow never used the mouse when presenting and it never occurred to me to click on it. I think a hint in the README.MD file would not be bad ;)

Thanks for the great tool!

### DIFF
--- a/README.md
+++ b/README.md
@@ -764,6 +764,8 @@ If you want to add a theme of your own see the instructions here: [/css/theme/RE
 
 reveal.js comes with a speaker notes plugin which can be used to present per-slide notes in a separate browser window. The notes window also gives you a preview of the next upcoming slide so it may be helpful even if you haven't written any notes. Press the 's' key on your keyboard to open the notes window.
 
+A speaker timer starts as soon as the speaker view is opened. You can reset it to 00:00:00 at any time by simply clicking/tapping on it.
+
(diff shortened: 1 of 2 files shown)


demoneaux:
LGTM.

Timothep:
@demoneaux: Dammit. It was intended to be a second PR with screenshot and all... sorry about that, I messed up with the Github UI. You're going to have to help me, how do I remove the second commit from the PR?

demoneaux:
It's kinda complicated, but perhaps for now you can just explain the rationale of the second PR?

Timothep:
Well, it's simply a different overview of the speaker notes (see below). The way it was before gave very little importance to the notes themselves (1/4th of the screen minus the "timer bar"). As a regular keynote-style speaker (quite minimalistic slides), I find myself wanting to see more notes and less slides.

This is certainly very subjective, that's why I didn't want to PR it right away :-/

revealspeaker

hakimel:
Thanks for pointing out that's hard to find. I think we should make the option to reset more obvious instead of having to explain it in the readme. Perhaps a small reset icon that's always visible?

Timothep:
That would sure work as well. But it's a feature nonetheless, why not add a note to the readme anyway?

hakimel:
These commits have been merged into the dev branch.

This includes the new speaker view layout as well, I think it makes a lot of sense to make more room for notes. Thanks!

PR #1532, opened 2016-03-16
update headjs to v1.0.3

booi:
Hi,

I noticed the version of headjs was very old and the new version seems to work. This also resolves a very odd issue of the mime-types gem detecting the old headjs file as text/html instead of javascript. I ran into this bug trying to use s3_website to deploy a static presentation website. The mime-type of head.min.js is detected incorrectly and modern browsers fail to execute it. 

Either way, reveal.js seems to work well with the new headjs.


hakimel:
Thanks! Looks like it's working well when I try from here too.

I merged your commit into the dev branch, please submit towards that branch directly next time.

Issue #855, opened 2014-03-30
Adapt parallax background size to resolution


Kabouik:
Hi,

I'm not a coder but I finally managed to make my first reveal.js presentation without GUI. A lot of pain, but a good satisfaction! 

Is it possible/planned to make parallax background stretch to adapt to screen resolution, with option for constraining aspect ratio?

I've got a presentation with a background parallax that looks good when viewed in 1080p, but most projectors are not 1080p and so they will display only the top left portion of the background, where there is nothing to see. If I reduce the parallax background size in the index.html, then I have to know in advance the resolution of the projector I will use, which is most likely impossible (otherwise, if I set a resolution lower than the projector's one, then the background will be too small and will appear repeated).

Thanks!

PR #1533, opened 2016-03-22
add parallax viewport to scale image depending on resolution

vincentbernat:
When providing a parallax image, a user may want to be able to influence
the image part that should be displayed. It would then provide an ideal
viewport for the image. If the viewport size matches the browser size,
nothing changes. However, if the browser size is smaller, the image will
be reduced to match the viewport and the reduction factor is also
applied to the steps. The ratio is kept and the scale factor is chosen
to "cover" the image (otherwise, parallax won't work).

This doesn't cover all possible situations, but it should be good enough
to handle resolution change without ratio change (1920×1080 → 1280×720)
or with a small ratio change (1920×1080 → 1024×768).

Fix: #855


rriemann:
This would solve my problem I guess. Is there a special reason why this has not been merged?

Kabouik:
Sorry for not answering earlier @vincentbernat, I did not see that you submitted a PR for the issue I opened a long time ago. 

I don't know why your PR was not merged, but I cloned it directly using `git clone https://github.com/vincentbernat/reveal.js.git -b feature/parallax-viewport reveal.js.parallax` to test it locally. It seems to work well, thanks. Unfortunately, unless I missed something, it doesn't allow setting the desired aspect ratio in advance. This is an issue because one does not always know which resolution and ratio the projector will use before a talk (and one of the selling points of reveal.js is to be portable), but it is still much better than the `master` parallax feature with no scaling at all. 

I tried using your `reveal.js` file instead of its counterpart in the `master` branch together with other `master` files, but unfortunately this breaks everything and nothing shows up in the slides. Would it be difficult to update the PR to make it work with the current `master` and benefit from the updated reveal.js **with** scaling parallax?

PR #1535, opened 2016-03-23
Topic README node v

LindsayBrown81:
I added the following note to the README.md on line 1008 (in the installation section, under full setup step 1 Install Node.js): Please note that reveal.js supports only the current version of Node; i.e., node 0.10 is no longer supported.


hakimel:
Thanks, it makes sense to note which version is required. I tweaked it and noted which version we do support rather than what we do not; 1efa52486c8712507b8a343b26b227e834da5fe6

LindsayBrown81:
Thank you. 

Lindsay Brown

PR #1537, opened 2016-03-24
make theme "simple" use white titles on dark backgrounds

vincentbernat:
This is similar to what is done in the "white" theme.


hakimel:
🙌

PR #1542, opened 2016-03-29
Update README.md

stephanmg:
Note about using percentage for data-background-size attribute.


hakimel:
Was looking at merging this but felt like the whole section on slide backgrounds was missing a lot of info so I rewrote it instead. The rewritten version includes a full description of all supported attributes, include the `data-background-position` which wasn't even mentioned before.

https://github.com/hakimel/reveal.js#slide-backgrounds

PR #1544, opened 2016-03-31
Updating package dependencies for newer versions

lirantal:
Issues with current packages:
* socket.io version is 6 months old and that version has insecure dependencies
* grunt-contrib-watch is 24 months old with insecure dependencies
* grunt-contrib-qunit is 12 months old with insecure dependencies


hakimel:
Thanks for the PR but you'll need to resubmit towards the 'dev' branch. There's recent changes to package.json that conflict with your changes.

Issue #1389, opened 2015-10-07
table of contents


ducin:
First of all, @hakimel, many thanks for your work. I do use slides.com a lot!

I can see that the readme.md file has lots of valuable content, but it is a little difficult to look through (due to its size and complexity). My suggestion is to introduce the **table of contents** just like the ones I did [here](https://github.com/json-schema-faker/json-schema-faker) and [here](https://github.com/swagger-api/swagger-codegen#table-of-contents).

But before I do that I'd like to know whether you're interested in having such thing. It's only aim is to improve readibility of the readme.

demoneaux:
@hakimel what do you think about this? I can send in a PR for this if you’re fine :)

hakimel:
I agree the readme is getting a bit unwieldy. I'm all for adding a TOC.

ducin:
I'm happy to create such table of contents. But I need you to accept it soon (otherwise it'll get outdated and might become useless at some point). Been waiting few moths for this reply :P. Hm?

demoneaux:
@ducin Feel free to send in a PR. We’ll try to merge it ASAP :)

demoneaux:
Closed by #1545.

PR #1545, opened 2016-04-02
Table of contents added in README.md

ducin:
This solves #1389. I cloned a fresh copy of master branch today.
Please merge it soon, hopefully, so that it doesn't get diverged from existing readme :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,64 @@
-# reveal.js [![Build Status](https://travis-ci.org/hakimel/reveal.js.svg?branch=master)](https://travis-ci.org/hakimel/reveal.js)
+# reveal.js
+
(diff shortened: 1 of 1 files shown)


ducin:
By the way, going through the TOC I saw how many great features are here in reveal.js :)
as I was browsing through the TOC, I could see that some of the paragraphs might be reordered, some might be wrapped in bigger sections.

In general, a big overview of the TOC might be a second step.

hakimel:
Thanks! This has been merged. The TOC grew pretty huge so I might actually exclude a few sub-topics.

ducin:
Thanks :) it's a good starting point to do some cleaning :)

PR #1550, opened 2016-04-08
Fixes #1549 , Chrome vs. Edge detection

### LINKED ISSUE #1549: Edge is accidentally using 'zoom'
Hi,

I noticed that in Windows 10 + Edge reveal.js is using 'zoom' property instead of 'transform: scale()'.

Edge UserAgent: 

*Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) 
Chrome/46.0.2486.0 Safari/537.36 Edge/13.10586*

That will match in to RegExp intented for Chrome it is found [here](https://github.com/hakimel/reveal.js/blob/master/js/reveal.js#L1592)
```js
/chrome/i.test( navigator.userAgent )
```

This is causing issue with Edge scaling at least in our project. When `zoom` property value is under `0.846` whole background image disappears, but if we use `transform: scale(_some_value_)` it works perfectly.

RopoMen:
Hi,

I excluded MS Edge from the Chrome detection and also optimized 'navigator.userAgent' access.

I also noticed that this Edge issue was probably fixed in 'dev' branch but I wanted to make sure that Edge is excluded from the Chrome, if there would be some issues in Edge with 'zoom' attribute [when upscaling](https://github.com/hakimel/reveal.js/blob/dev/js/reveal.js#L1611).

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -32,6 +32,7 @@
 		HORIZONTAL_SLIDES_SELECTOR = '.slides>section',
 		VERTICAL_SLIDES_SELECTOR = '.slides>section.present>section',
 		HOME_SLIDE_SELECTOR = '.slides>section:first-of-type',
+		UA = navigator.userAgent,
 
 		// Configuration defaults, can be overridden at initialization time
 		config = {
@@ -197,6 +198,9 @@
 		// Client is a mobile device, see #checkCapabilities()
 		isMobileDevice,
 
+		// Client is a desktop Chrome, see #checkCapabilities()
+		isChrome,
+
 		// Throttles mouse wheel navigation
 		lastMouseWheelStep = 0,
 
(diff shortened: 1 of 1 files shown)


RopoMen:
I have also one question, when I can expect next release to be out?

hakimel:
Thanks! I don't have Edge installed anywhere that I can test but the diff looks good and Chrome still works so I went ahead and merged.

RopoMen:
Hi,
Thanks, I have used [modern.ie virtualboxes](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/windows/) for testing.

btw. about that new release, is it coming out soon? :)

hakimel:
Thanks, I use those too but need downloaded them ages ago before Edge. Not
sure when the release will be. Probably soon :)

PR #1552, opened 2016-04-09
README: add encoding to html example

xrmx:
It avoids a huge red warning in FF console when loaded and may
save some hair pulling later.


hakimel:
Thanks but I want to keep that example small to highlight the essential components of reveal.js. We define charsets in the standard index.html file though.

PR #1556, opened 2016-04-11
Remove duplicated webkitRequestFullScreen method

dominhhai:
Removed the duplicated `webkitRequestFullScreen` method in `enterFullscreen`

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -1977,11 +1977,10 @@
 		var element = document.body;
 
 		// Check which implementation is available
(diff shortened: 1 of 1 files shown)


hakimel:
Those two aren't identical – some versions of webkit used lowercase **s** in "full**s**creen" where others used uppercase "full**S**creen". Has this changed?

dominhhai:
ooh, I didn't notice that. Thank you & Sorry!

PR #1561, opened 2016-04-16
Add Dockerfile to run reveal.js in docker

FuriKuri:
I added a Dockerfile with setup, so you can easily run reveal.js within
a docker container. This also makes it easier to start reveal.js on a
remote server.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,18 @@
+FROM node:argon
+
+RUN mkdir -p /usr/src/app
+
+COPY package.json /usr/src/app
+COPY Gruntfile.js /usr/src/app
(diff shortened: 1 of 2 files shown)


demoneaux:
I think this is too specific for the general Reveal.js project. Also, running the `grunt` script to serve the presentation isn't recommended for production, and is only meant for development. @hakimel opinions?

hakimel:
Thanks for the PR @FuriKuri. I agree with @demoneaux. I don't think reveal.js should be concerned with where or how a presentation is deployed/hosted. There's so many different environments you might want to run it in and we can't maintain configs for all of them in this repo.

PR #1563, opened 2016-04-16
display pacing advice based on slide timings

aspiers:
Display advice on whether the current pace of the presentation is on track for the right timing, and if not, whether the presenter should speed up or has the luxury of slowing down.

By default this assumes 120 seconds per slide, which can be a reasonable rule of thumb, but this can be configured by the `defaultTiming` parameter in the `Reveal` configuration block, and also per slide `<section>` by setting the `data-timing` attribute.  Both values are the number of seconds.

Here is a screenshot:

pacing

The pacing advice changes red when you are behind, and amber when ahead.


aspiers:
Submitted against wrong branch - resubmitting against `dev`.

PR #1564, opened 2016-04-16
display pacing advice based on slide timings

aspiers:
Display advice on whether the current pace of the presentation is on track for the right timing, and if not, whether the presenter should speed up or has the luxury of slowing down.

By default this assumes 120 seconds per slide, which can be a reasonable rule of thumb, but this can be configured by the `defaultTiming` parameter in the `Reveal` configuration block, and also per slide `<section>` by setting the `data-timing` attribute.  Both values are the number of seconds.

Here is a screenshot:

pacing

The pacing advice changes red when you are behind, and blue when ahead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,10 +494,11 @@ Reveal.getScale();
 Reveal.getPreviousSlide();
 Reveal.getCurrentSlide();
 
-Reveal.getIndices(); // { h: 0, v: 0 } }
+Reveal.getIndices();        // { h: 0, v: 0 } }
 Reveal.getPastSlideCount();
-Reveal.getProgress(); // (0 == first slide, 1 == last slide)
-Reveal.getTotalSlides(); // total number of slides
+Reveal.getProgress();       // (0 == first slide, 1 == last slide)
+Reveal.getSlides();         // Array of all slides
+Reveal.getTotalSlides();    // total number of slides
 
 // Returns the speaker notes for the current slide
 Reveal.getSlideNotes();
--- a/README.md
+++ b/README.md
@@ -497,7 +497,7 @@ Reveal.getCurrentSlide();
 Reveal.getIndices(); // { h: 0, v: 0 } }
 Reveal.getPastSlideCount();
 Reveal.getProgress(); // (0 == first slide, 1 == last slide)
-Reveal.getTotalSlides();
+Reveal.getTotalSlides(); // total number of slides
 
 // Returns the speaker notes for the current slide
 Reveal.getSlideNotes();
--- a/README.md
+++ b/README.md
@@ -495,6 +495,7 @@ Reveal.getPreviousSlide();
 Reveal.getCurrentSlide();
 
 Reveal.getIndices(); // { h: 0, v: 0 } }
+Reveal.getPastSlideCount();
 Reveal.getProgress(); // (0 == first slide, 1 == last slide)
 Reveal.getTotalSlides();
 
--- a/README.md
+++ b/README.md
@@ -495,7 +495,7 @@ Reveal.getPreviousSlide();
 Reveal.getCurrentSlide();
 
 Reveal.getIndices(); // { h: 0, v: 0 } }
-Reveal.getProgress(); // 0-1
+Reveal.getProgress(); // (0 == first slide, 1 == last slide)
 Reveal.getTotalSlides();
 
 // Returns the speaker notes for the current slide
--- a/README.md
+++ b/README.md
@@ -187,6 +187,9 @@ Reveal.initialize({
 	// Display a presentation progress bar
 	progress: true,
 
+	// Set default timing of 2 minutes per slide
+	defaultTiming: 120,
+
 	// Display the page number of the current slide
 	slideNumber: false,
 
@@ -967,12 +970,15 @@ Notes are only visible to the speaker inside of the speaker view. If you wish to
 
 When `showNotes` is enabled notes are also included when you [export to PDF](https://github.com/hakimel/reveal.js#pdf-export). By default, notes are printed in a semi-transparent box on top of the slide. If you'd rather print them on a separate page after the slide, set `showNotes: "separate-page"`.
 
-#### Speaker notes clock and timer
+#### Speaker notes clock and timers
 
 The speaker notes window will also show:
 
 - Time elapsed since the beginning of the presentation.  If you hover the mouse above this section, a timer reset button will appear.
 - Current wall-clock time
+- (Optionally) a pacing timer which indicates whether the current pace of the presentation is on track for the right timing (shown in green), and if not, whether the presenter should speed up (shown in red) or has the luxury of slowing down (blue).
+
+The pacing timer can be enabled by configuring by the `defaultTiming` parameter in the `Reveal` configuration block, which specifies the number of seconds per slide.  120 can be a reasonable rule of thumb.  Timings can also be given per slide `<section>` by setting the `data-timing` attribute.  Both values are in numbers of seconds.
 
 
 ## Server Side Speaker Notes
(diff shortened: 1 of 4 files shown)


fghaas:
So I had a look at this and it's certainly useful, if not *quite* as useful as it could be. A few issues:

- What's a typical reason for not finishing a slide on time? Taking a question. Would you want to discourage taking questions? I hope not. So maybe you'd want to be able to to pause the countdown while taking a question, and intuitively you might want to click on the timer. Currently that resets the whole talk timer to zero, meaning your pacing now assumes that you have an additional buffer equal to the combined time budget of all prior slides. So that's a missing feature (which is fine, can always be added later) combined with somewhat counter-intuitive and illogical behavior (which really should be fixed).

- What do you typically do to account for that in your time budget? You plan a reserve at the end of the talk. Defining a reserve is something you may want to consider, as in "you're over time on *this* particular slide, but you've only used 30 seconds of a 5-minute buffer".

- Setting the default to 120s is problematic. The "2 minutes per slide" rule gets repeated just as often as the "no more than 7 bullets per slide" rule, and both are equally wrong — the time you spend on a slide is inherently flexible based on a number of factors, just as 7 bullets is way too many. I'd suggest having the default be set to 0, in which case the pacing information should also be suppressed in the speaker console.

aspiers:
> Would you want to discourage taking questions?

I think there are two approaches to talks:

1. Insist that people hold their questions until the end (important with large audiences)
2. Welcome questions at any point (better with smaller audiences, at less formal events)

Clearly, a strict timing plan will not work as well with the latter, but it should with the former.  But having said that ...

>  So that's a missing feature (which is fine, can always be added later) combined with somewhat counter-intuitive and illogical behavior (which really should be fixed).

Congratulations for correctly predicting my next plans for this feature :-)  I was thinking of adding `+1` and `+5` minute buttons to add more time - not just for questions, but also sometimes the event organizer decides to give you more time during the middle of the talk.  But your pause button idea is a great idea - probably even better.  Maybe we can have both.

> Defining a reserve is something you may want to consider

Another excellent idea!  Yes - allocating a 5 min reserve would mean that it could be used up accumulatively at any points during the talk.  Saving it to the end is ideal, but having flexibility to take questions earlier is even better.  Question: should a pause button always toggle countdown between the main timer and the reserve timer?  Or sometimes do you really want to pause everything?  That would have the effect of delaying the end of the presentation, of course.

> the time you spend on a slide is inherently flexible based on a number of factors

This is of course true.  However in my experience, when *averaged* out over a *whole* presentation, 2 minutes per slide is startlingly accurate as rules of thumb go, for *my personal* presentation style.  Your style tends to use far sparser slides, so I doubt it would work as well for you.  That said, you may well be right that it makes more sense for this feature to be suppressed by default.

aspiers:
The other problem I have noticed is that generally we don't want the timer to start until the 2nd slide or so, e.g. if the 1st slide is a QR code or other cover slide which is to be shown whilst waiting for the audience to get settled.  Suggestions very welcome for how best to handle this.

fghaas:
I'd say for any slide that doesn't count toward the time budget, the onus is on the speaker to set `data‑timing="0"`, which should be interpreted as "this slide has no time budget," rather than "this slide has a time budget of zero, except when massless and traveling at _c_ which would make the time budget ∞".

aspiers:
I was thinking `"-1"` might be a more apt value, but it doesn't matter much.  The difficult bit is figuring out how to pause the timer when it's on that slide.  Would you say this would only happen for slides at the beginning?

fghaas:
I'd say if a slide has `data‑timing="0"` (or `"-1"`, whichever you prefer), then just pause the pacing timer as soon as you get to that slide. For the overall time budget count it as zero. Doesn't matter whether the slide is at the beginning or midway through.

Yes of course you should also have an option to _pause_ the timer, rather than only be able to reset it, but reveal currently doesn't have that for the "elapsed time" timer either. So that's icing on the cake.

aspiers:
Just thought of a second use case for "pause" slides: a long session (e.g. hands-on training) where you want to allocate flexible breaks at various points.  So yes, something like `data-timing="pause"` seems to make sense.

Current thinking:

- add a new `totalIgnoredTime` variable
- add a `pauseStart` variable which is normally `null`, but gets set to wall-clock time when either a slide with `data-timing="pause"` is entered, or a `pause` button is manually clicked
- when a "pause" slide is exited or the `pause` button is released, the elapsed time since `pauseStart` is added to `totalIgnoredTime` and then `pauseStart` set back to `null`
- calculation of time left for the current slide adds `totalIgnoredTime` to the time available
- clicking `+1 min` or `+5 min` buttons would add to `totalIgnoredTime` accordingly

I still haven't figured out the best UI regarding a "reserve" timer though.  Intuitively, allowing setting of a reserve timer to (say) 5 minutes as a buffer for unplanned questions makes sense.  However if the presenter presses the Pause button, that could be either

a) to answer a question, in which case it should start eating away at the reserve, or 
b) it could be for the kind of break mentioned above, in which case everything should be paused.

So how to distinguish between these two types of pause in the UI?

Would also be nice to show:

- how much of reserve is remaining
- current ETA for session finish
- estimate of how long the remaining material should take

whilst somehow retaining a simple and intuitive UI.

fghaas:
You're burrowing deep in a rabbit hole here. I'd say keep your changes as they are for now, except setting the default to 0 and not showing the pacing timer if the pacing value for the current slide (either via `data-timing` or by default) is <=0, and get this merged. *Then* add bells and whistles.

Or, drop this PR and start your own plugin.

aspiers:
Sure, those enhancements can come in a separate PR to this one.

It definitely doesn't make sense to start another plugin though, because it's inextricably linked with the existing built-in `notes` plugin.

fghaas:
Point taken. (I've never really understood how _some_ plugins live in the main repo and others don't, but that's a different story.)

aspiers:
FWIW this plugin helped me and a colleague successfully deliver > 50 slides and 2 demo videos within 36 minutes ;-)  https://www.openstack.org/videos/video/high-availability-for-pets-and-hypervisors-state-of-the-nation

fghaas:
@hakimel This looks like a very useful feature to me at this point; would you consider merging it?

hakimel:
This is really nice! Thanks @aspiers.

I'd be happy to merge this with one change; can you make it so that this is disabled by default? Don't want to make assumptions about how long someone wants to spend per slide (we can suggest 2 mins in the readme though), and I think seeing a countdown there without having opted into it is a bit surprising.

aspiers:
@hakimel Sure, I can do that.  Do you think it would be reasonable if it was enabled by the presence of the `defaultTiming` configuration parameter?

hakimel:
Thanks! Yea, I think basing it on the presence of that config value makes
sense. Matches how other config values work as well.

aspiers:
@hakimel I have just rebased this branch and made the feature optional as requested, so that it's only activated when `defaultTimings` is configured.  /cc @fghaas

hakimel:
Thanks for the update – this has been merged.

Made a minor tweak; instead of appending "to finish current slide" to the clock where the text is so large I added it into the "PACING" label.

I wonder if you ever want to have both the elapsed time and pacing... Feels like if you opt into pacing the elapsed time counter is only visual clutter that makes it harder to follow your pacing. Maybe we should hide the elapsed time when pacing is turned on?

fghaas:
Awesome! Thanks for merging this @hakimel, and congrats to @aspiers for getting it landed.

aspiers:
> I wonder if you ever want to have both the elapsed time and pacing... Feels like if you opt into pacing the elapsed time counter is only visual clutter that makes it harder to follow your pacing. Maybe we should hide the elapsed time when pacing is turned on?

Personally I prefer to have it there, and the colouring distinguishes the pacing timer from the previous one - but we could do that, or make it optional.

bdarcus:
> I wonder if you ever want to have both the elapsed time and pacing... Feels like if you opt into pacing the elapsed time counter is only visual clutter that makes it harder to follow your pacing. Maybe we should hide the elapsed time when pacing is turned on?

I agree with your impulse on this, and your conclusion.

PR #1572, opened 2016-04-25
More natural zooming on block level elements

jesstelford:
Switching a `display: block` element to `display: inline-block` allows calculating the bounds based on the contents of the div rather than the entire container (which is often `width: 100%`).

This provides a much more natural zoom, especially for paragraphs and code examples.


hakimel:
Looks good, thanks!

PR #1575, opened 2016-04-27
Add selection colour for Mozilla Firefox

aniqah:
On Mozilla Firefox, the default selection/highlight colour is used when selecting slide content (text, images, etc). This change makes it so that the colour defined for other browsers in each theme will be used on Firefox as well. No more clashing colours! :)

### DIFF
--- a/css/theme/beige.css
+++ b/css/theme/beige.css
@@ -29,6 +29,11 @@ body {
   background: rgba(79, 64, 28, 0.99);
   text-shadow: none; }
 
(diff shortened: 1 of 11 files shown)


hakimel:
Cheers! Fix makes sense, but you added the selection colors to the compiled theme CSS file so it'll be overwritten whenever the SCSS compiles. I went ahead and added it to the theme.scss template so that it'll be automatically generated for each theme.

PR #1576, opened 2016-04-28
add doclets.io config

lipp:
Adds documentation to [doclets.io](https://doclets.io) service. Future tags and master branch will be published automatically.

[Preview generated with my account](https://doclets.io/lipp/reveal.js/add-to-doclets)

@hakimel  If you like this, you are very welcomed to login to https://doclets.io with GitHub OAuth. If you don't like it, I'd be happy to hear what you don't like about it :)

Steps required (3min):
1) @hakimel   logs in to doclets.io
2) @hakimel   adds reveal.js
3) someone merges this PR

### DIFF
--- a/.doclets.yml
+++ b/.doclets.yml
@@ -1,0 +1,3 @@
+dir: js
+articles:
+  - Readme: README.md


demoneaux:
@hakimel any opinions on this docs service, or do you think the readme is enough?

hakimel:
Thanks but I don't think this is useful right now. Most of what's in the doc is internal, with limited detail in the comments. I'd rather stick to the README for now.

PR #1577, opened 2016-05-01
add refused-to-display tips in link-preview-overlay

Huxpro:
PreviewLinks is a super convenient feature while many of websites are refused to display their webpages in a frame due to their policy. Some set Content Security Policy "frame-ancestors 'none'" and others set 'X-Frame-Options'.

These kind of errors don't trigger `iframe.onerror` which is a big mis-implementation of browsers in my views. So we have to work around it with some tricks and my proposal is to place a `z-index: -1` element under the iframe, which would be only saw if the iframe display nothing (and it mostly due to sort of 'refused-to-display' policy)

### DIFF
--- a/css/reveal.scss
+++ b/css/reveal.scss
@@ -1229,6 +1229,16 @@ html:-moz-full-screen-ancestor {
 		visibility: visible;
 	}
 
+	.reveal .overlay.overlay-preview.loaded .viewport-inner  {
+		position: absolute;
+		z-index: -1;
+		left: 0;
+		top: 60px;
+		width: 100%;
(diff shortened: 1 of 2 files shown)


demoneaux:
@hakimel any opinions? I think this might be quite useful.

hakimel:
This is a good improvement. This message will appear whenever an iframe is unable to load, so I have changed the message a little bit. I really wish there was a way to explicitly detect when an iframe fails to load because of `x-frame-options`. 

The error message was visible briefly when an iframe loaded successfully as well but I've added a transition to fix that.

PR #1585, opened 2016-05-09
set/remove @disabled on navigation buttons

jasonkiss:
Keeps disabled buttons out of kbd tab order and indicates to assistive tech like screen readers that button is disabled. Otherwise buttons that aren’t enabled remain in kbd Tab order and screen readers announce them as buttons, but they don’t work.


hakimel:
Great stuff. Thanks again!

PR #1586, opened 2016-05-09
make speaker notes keyboard accessible

jasonkiss:
Place div.speaker-notes in default kbd Tab order, and when focused,
prevent slide navigation with up/down arrows, allowing up/down arrow
keys to scroll div.speaker-notes. Otherwise, sighted keyboard-only users cannot scroll and view speaker notes that are longer than height of div.speaker-notes.


hakimel:
Good improvement, thanks!

PR #1595, opened 2016-05-19
Add minor imporovements to the PDF export instructions

tristansokol:
I had a couple minutes of fiddling to get the PDF export to work that I think can be saved with a minor language adjustment in the instructions, Namely highlighting that the PDF export capabilities are only available when viewing the presentation from a web server.


hakimel:
Thanks!

PR #1598, opened 2016-05-21
Removing duplicated "position" property at pdf.css

tiagorg:
Found that by running css-lint on the code


hakimel:
Thanks, merged into dev https://github.com/hakimel/reveal.js/commits/dev

Issue #1262, opened 2015-06-12
Per slide vertical centering


hadim:
Most of my slide are not vertically centered (I use `center: false,` option). How can I ask reveal.js to vertically center only one slide (by JS or CSS).

Thank you

tpgmartin:
Hey @hadim this should be possible by adding the class "center" to any slide you want vertically centred.

hadim:
Thank you +1

jdehaan:
+1 Good idea!

denispeyrusaubes:
How can you add the "center" class to a slide in your markdown code ?

---
## Title

`To be vertically centered`

demoneaux:
You can see [this example from the readme](https://github.com/hakimel/reveal.js#slide-attributes):

```html
<section data-markdown>
    <script type="text/template">
    <!-- .slide: class="center" -->
        Markdown content
    </script>
</section>
```

Or just add the class directly as well:

```html
<section data-markdown class="center">
    <script type="text/template">
        Markdown content
    </script>
</section>
```

denispeyrusaubes:
Thank for the response... 
but I need to apply this style to some specific slides, and my slides are divided in chapters I specific md files.
One slide is delimited by ```---``` from the other
so I don't understand how to proceed in my case.

My index.html links all chapters using:

```html
<section>
				<section data-markdown="/myfolder/mychapter.md" data-separator="^\n---\n" data-vertical="^\n\n">
					Markdown body of a slide here ...
				</section>
			</section>
```

demoneaux:
You can try placing the `<!-- .slide: class="center" -->` after each `---` for each slide.

denispeyrusaubes:
Strange, I've placed this code immediately after ```---```, and the generated html contains

```html
<section data-markdown="" data-markdown-parsed="true" class="center present" style="top: 0px; display: block;"><!-- -->
<h2 id="le-jeu-de-test">Le jeu de test</h2>
<p>test</p>
</section>
```

But my text isn't vertically centered...

demoneaux:
The slide should be vertically centred, not too sure how to help you with that anymore.

denispeyrusaubes:
is ```center``` a standard css style ? 
I can't find a specific css rule defined in provided css files...

demoneaux:
[...]

PR #1608, opened 2016-06-01
Slide-specific vertical centering override

albinotonnina:
When a specific section has to be shown not vertically aligned to the center.
`    <section data-vertical-align-top>
`

### DIFF
--- a/css/reveal.scss
+++ b/css/reveal.scss
@@ -478,6 +478,10 @@ html:-moz-full-screen-ancestor {
 	opacity: 0;
 }
 
+/* Slide-specific vertical centering override */
+.reveal .slides section[data-vertical-align-top]{
+	top: 0 !important;
+}
 
 /*********************************************
  * Mixins for readability of transitions


elvin100:
Thank you for this! Can you explain me where exactly to put this? When I'm trying it - it is breaking my slides.

mshibanami:
+1.
This PR looks abandoned, but still attractive for me.

bnjmnt4n:
This seems like a good idea, although the name of the property might need to be changed. Currently we enable vertical centering if the `center: true` configuration is set or the slide has a class of `"center"`. Perhaps looking at the class of `"no-center"` might be a better idea for consistency (or maybe a separate `data` attribute).

utdrmac:
Came here, 2 years later after the most recent post, looking for the ability to have all my slides aligned at the top by default. Still no traction here? +1 to this feature/ability.

edbennett:
Another few years later, I was looking for how to do this; the top result on DuckDuckGo was #1262, which does the opposite—but you wouldn't want to have to specify `center` on every slide apart from the one that you want to top-align.

PR #1611, opened 2016-06-04
showHelp exposed

rsp2k:


hakimel:
Thanks, I agree that could be useful in the API. There were a few issues with the PR though (it contains an unrelated video controls change and an unused argument for `showHelp`) so I added it myself in ca92d22adcf2a5a6c960b2d4a6bdadd13b12385b

Issue #1614, opened 2016-06-08
Markdown Example missing Script elements.


eddiewebb:
Slide 6 in demo omits the required <script template tags.

eddiewebb:
Submitted PR #1615 to resolve this.  I think I followed all guidance in contributing.md

demoneaux:
The `<script>` template tags are not required for all usage of Markdown, only if HTML tags are still used within the Markdown, to prevent browsers from parsing it as HTML before Markdown gets parsed.

PR #1615, opened 2016-06-08
Addresses issue #1614 by adding script element to example

eddiewebb:
Addresses #1614 by adding `<script>` elements to quoted markdown.
Also added html tag to markdown syntax block.


demoneaux:
See https://github.com/hakimel/reveal.js/issues/1614#issuecomment-224667986.

PR #1628, opened 2016-06-23
Detail markdown plugin attributes & defaults

davidjb:
Expands the details already given by highlighting the defaults for all attributes and clarifying what's required.


hakimel:
Looks great, thanks!

PR #1632, opened 2016-06-28
Add Transition: Appear

Bramas:
Add new transition: Appear
It between a vertical slide and a fade in.

(And don't know if the name is good)

### DIFF
--- a/css/reveal.css
+++ b/css/reveal.css
@@ -743,6 +743,87 @@ body {
           transition: none; }
 
 /*********************************************
+ * APPEARING TRANSITION
+ *********************************************/
+.reveal .slides > section[data-transition=appear].past,
+.reveal .slides > section[data-transition~=appear-out].past,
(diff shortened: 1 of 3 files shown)


hakimel:
This looks nice. "Appear" doesn't really indicate what the transition looks like, maybe "slide-up" is a better name?

hakimel:
Also, there are a lot of complex selectors related to timing. Is there a way we can remove or simplify those?

Bramas:
I don't know. The selector are quiet similar to the other transition, except for the fact that I had to force the speed to be fixed because I felt that it was wrong otherwise. 

I use this transition for all my presentations and I felt it could be usefull to others but I didn't had time to optimize it.

I took the name from the same effect in the keynote software. It's really close the slide transition but more sober, so maybe fadein-slide, or slide-in.

btw, thank you for this great projet, it's the best 👍 !!

PR #1636, opened 2016-07-04
For a fragment: allow to show a separate note defined in it

traff:
When a slide has several fragments it could be convenient to define a note for each of them. In this case we need to show only this specific note defined in a fragment and not others. General note of a slide shouldn't be also shown, as a more specific one should have greater relevance in this case.

### DIFF
--- a/plugin/notes/notes.js
+++ b/plugin/notes/notes.js
@@ -50,7 +50,7 @@ var RevealNotes = (function() {
 		/**
 		 * Posts the current slide data to the notes window
 		 */
-		function post() {
+		function post(event) {
 
 			var slideElement = Reveal.getCurrentSlide(),
 				notesElement = slideElement.querySelector( 'aside.notes' );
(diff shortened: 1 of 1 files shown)


hakimel:
Great idea 🙌

hakimel:
^ made a few changes to this, to make it work correctly when stepping to and from a slide with multiple fragment notes. Also added support for notes in an attribute `<div class="fragment" data-notes="Notes here">`.

midiisc:
Hi. I was trying out this feature but it doesnt work in the sense when I added the <aside> tags in different fragments, only the first fragment notes were displayed and others were ignored. Is it something wrong in Im doing or is the framework still looking for first occurance of <aside> inside section tag and ignoring all subsequent ones. 

Also, It will be nice if possibility for timed highlighting of speaker notes (change in color or bold text etc.) can be implemented. As it helps in adhering to a timed script for a presentation. More specifically, if there is a video in the presentation,. voice over becomes much more easier with this. Im sorry if this is something which already exists but I couldnt locate it.

Best regards

eric-wieser:
This doesn't work optimally if there are multiple current fragments - it only pulls the notes for the first one.

appfinder:
Hello,  can someone please give an example how achieve this effect, i couldn't(t find any in documentation

PR #1644, opened 2016-07-21
 	Turn off preview links locally

rajgoel:
Currently, the global configuration can be used to turn preview links on or off. For `previewLinks: false` it is possible to locally turn on preview links by `<a href="..." data-preview-link>Link</a>`.

This pull request allows to turn off preview links for `previewLinks: true` by `<a href="..." data-preview-link=false>Link</a>`.

This is particularly useful if you want to have preview links in general, but some pages do not allow to be opened in an iframe.


hakimel:
Makes sense 👍

PR #1649, opened 2016-07-30
Workaround rendering glitch in Firefox 47+

jackycute:
Including slides, overview been cutoff and text became blurry

### DIFF
--- a/css/reveal.scss
+++ b/css/reveal.scss
@@ -394,6 +394,7 @@ body {
 	text-align: center;
 	perspective: 600px;
 	perspective-origin: 50% 40%;
+	transform-style: preserve-3d;
 }
 
 .reveal .slides>section {
@@ -408,7 +409,7 @@ body {
 	padding: 20px 0px;
 
 	z-index: 10;
-	transform-style: preserve-3d;
+	transform-style: flat;
 	transition: transform-origin 800ms cubic-bezier(0.260, 0.860, 0.440, 0.985),
 				transform 800ms cubic-bezier(0.260, 0.860, 0.440, 0.985),
 				visibility 800ms cubic-bezier(0.260, 0.860, 0.440, 0.985),
@@ -842,6 +843,7 @@ body {
 	top: 0;
 	left: 0;
 	perspective: 600px;
+	transform-style: preserve-3d;
 }
 	.reveal .slide-background {
 		display: none;
@@ -1011,6 +1013,7 @@ body {
 .reveal.overview {
 	perspective-origin: 50% 50%;
 	perspective: 700px;
+	transform-style: preserve-3d;
 
 	.slides section {
 		height: 100%;


potch:
Hey there- I'm from Mozilla and would like to help figure out what the bug is on our end. Do you have a reference URL where I can see slides with this issue? Also, what version of Firefox and OS are you seeing this in? I see you mention 47+.

jackycute:
Hi @potch, thanks for coming here.
For the blurry text, you could directly visit http://lab.hakim.se/reveal-js.
I'm running on both WIndows 8.1 and Mac OS X 10.11 with Firefox 47+

But seems the overview and cutoff problem not showing in above link.
It might be the nest DOM and styles cause that problem, because I'm using plugin to covert markdown to slides.

Try this https://hackmd.io/p/slide-example
and disable below styles using DevTool and you might able to reproduce this issue.
```css
.reveal .slides, .reveal .backgrounds, .reveal.overview {
    transform-style: preserve-3d;
}

.reveal .slides > section,
.reveal .slides > section > section {
    transform-style: flat;
}
```

Thanks!

rajgoel:
@potch 

Here's another example http://courses.telematique.eu/mathematics/?session=3 

The glitch where the right part is cutoff appears when moving horizontally (left/right), vertical scroll (up/down) works. Using FF48.0 on Ubuntu 14.04

Please, note that I might be change transition mode to "none" in a few days because I will release the slides to students and don't want the glitch to be there then.

hakimel:
Appreciate the PR. I had already switched to `transform-style: flat` for all non-3d transitions in the dev branch for a separate issue. The 3d transitions like concave and convex still used `preserve-3d` since they need it for the vertical 3d transitions to work.

I've revised this in 0b9233cf4a1a53297356cbc834c2240a92d20545 so that preserve-3d is only used specifically for vertical stacks, and only when a 3d-transition is being used. This latest version is available to test on http://lab.hakim.se/reveal-js/

jackycute:
Not only the transition, the overview need to set that `transfrom-style: flat`.
I tried 0b9233c in FF 47, 48, 49 on OS X 10.11.6, the overview looks like below:

hakimel:
Gotcha. Not sure why but I'm not seeing that overview error in FF 48 on OS X 10.11.4. Anything special required to trigger it?

jackycute:
No, it's just very normal open and press.
I also tested FF48 on Windows 8.1, the overview looks good except it has been cutoff at the first load.

hakimel:
Alright, was able to see the overview rendering issue in FF now too. Please give it a try at http://lab.hakim.se/reveal-js when you have a chance.

Let me know if you still see the problem and I'll reopen.

jackycute:
The cutoff on the first show up of overview in FF 49 on Windows persist.

It might be a FF rendering bug or what.

OS X runs everything prefect! No glitch, no stretch, good job!
Thank you for making such awesome project!

hakimel:
Damn. Thanks for testing. I was testing OS X too and things are working well there.

hakimel:
Tested FF on Windows here and it seemed okay. Not sure under what exact circumstance the clipping in the screenshot above is occurring, but if anyone finds any new info please file that in a new issue.

PR #1651, opened 2016-08-03
add dependency status badge and update dependencies 

bufferoverflow:


bufferoverflow:
@hakimel Do you see any option to get this merged?

bufferoverflow:
Would be cool to have the dependencies updated, also from a security perspective...

bufferoverflow:
dependencies are update so this is obsolete

PR #1660, opened 2016-08-11
Base to root

ClintH:
Reveal didn't work when using the 'base' option to locate the slideshow source files at a different location to where the Reveal server is located. It seemed to be because the 'base' option is also used by Grunt.

This patch changes the name of the option to 'root', and further, allows for several root paths to be used. This means you can locate your content for a presentation in one directory, and have the Reveal connect server fallback to loading themes and JS etc from it's primary location.


hakimel:
Merged, thanks!

Issue #1656, opened 2016-08-05
Reveal.js Fails to respect aria-hidden=true in content


malcalevak:
I appreciate that Reveal.js actually has a modicum of accessibility support built into it, however I discovered that the copying of screen content into the aria-status-div ironically doesn't take into account content that's been set to aria-hidden="true" nor content that's display: 'none' for that matter.
I've hacked in code to remedy the aria-hidden capability, but it uses jQuery, so it's not really a solution I could submit. Detecting display status I imagine would be substantially harder, given that it's not likely set inline, but ultimately I believe screen readers aren't supposed to see that content either. It wouldn't surprise me if there's even more accessibility related functionality that should be taken into account, but nothing quite as major as respecting aria-hidden="true"

hakimel:
This was fixed by @Sonaryr in #1665 and will make its way out with the next release.

PR #1665, opened 2016-08-18
Hide invisible status text

Sonaryr:
this pull request will no longer add aria-hidden="true" and display: none elements to the status div text.

Solves https://github.com/hakimel/reveal.js/issues/1656


hakimel:
This is a great fix. Thanks Maarten!

PR #1666, opened 2016-08-19
ignore delta=0 mousewheel event

### LINKED ISSUE #1607: Mousewheel skips backward
Using a trackpad (Mac), I get sent back one slide as soon as two fingers touch.

It would also be nice to be able to change the threshold for initiating a slide change.

gera2ld:
fix #1607


hakimel:
Good stuff, thanks!

PR #1677, opened 2016-08-26
Add note to README about how to disable scaling

mispy:
It took me a little while to figure out that this was possible, might be useful to have it in the README for others :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,20 @@ Reveal.initialize({
 });
 ```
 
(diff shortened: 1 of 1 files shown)


RopoMen:
I would recommend to make this 

```js
Reveal.initialize({
  disableZoom: true
});
```

and do the magic inside Reveal js side and place those settings as your PR shows. yes, this needs some extra work to move those things inside the reveal.js file + test + doc update.

hakimel:
Makes sense to cover this in the readme, has been a pretty common question.

PR #1679, opened 2016-08-29
Fixed slide function in the case no slides exist (Issue #1678)

### LINKED ISSUE #1678: Reveal.initialize breaks on empty slides
In the case slides are empty `Reveal.initialize` breaks on slide() function, initialize() -> readURL() -> slide() tries to set to the first slide (which doesn't exist)

mprinc:
In the process of inviting Reveal.initialize()

In the case slides are empty `Reveal.initialize` breaks on slide() function, initialize() -> readURL() -> slide() tries to set to the first slide (which doesn't exist)

This fixes slide() function by checking if slide exist before accessing it
Closes #1678

### DIFF
--- a/index.html
+++ b/index.html
@@ -26,6 +26,10 @@
 			<div class="slides">
 				<section>Slide 1</section>
 				<section>Slide 2</section>
+				<section>
+					 <section>Vertical Slide 1</section>
+					 <section>Vertical Slide 2</section>
+			 </section>
 			</div>
 		</div>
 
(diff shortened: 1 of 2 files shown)


hakimel:
Thanks for the PR but I don't want to make those changes to the demo presentation. Also adjusted the fix a little bit so that the whole slide function aborts when there are no slides. 0b6585bd44fe86344408d9d3ab8d3c17e254738a

PR #1701, opened 2016-09-14
Use partials for slides instead of single `index.html`

hbrysiewicz:


abea:
You'd be able to have multiple slides in each partial, right? I like this idea.

hbrysiewicz:
Yes! You would be able to break it up into whatever arbitrary sections you would like. I recently had to collaborate on a large presentation for training and having several devs work on the same `index.html` file was causing a lot of headaches and merge conflicts. Not sure why tests are failing but I will be able to look into it later.

PR #1716, opened 2016-10-01
autoSlide duration with playbackRate

chadmorrow:
The autoSlide duration of media elements with data-autoplay now takes the playbackRate of said media element into account when setting the duration so that autoSlide and the media element stay in sync.


hakimel:
Good fix, thanks!

PR #1717, opened 2016-10-01
Logitech R400 on Ubuntu blackout button support

jroper:
On Ubuntu 16.04, the Logitech R400's blackout button gets mapped to 'v' (keyCode 86). I don't know why this is, but since nothing else is currently using 'v', this allows it to be supported out of the box.


hakimel:
Thanks, merged into the dev branch https://github.com/hakimel/reveal.js/commits/dev

PR #1721, opened 2016-10-10
Avoiding a crash when there is zero <section>

Lucas-C:
Minimal bug reproduction code:

```
<!doctype html>
<html>
    <body>
        <div class="reveal">
            <div class="slides">
            </div>
        </div>
        <script src="js/reveal.js"></script>
        <script>Reveal.initialize({});</script>
    </body>
</html>
```


hakimel:
Thanks but this was fixed a while back in 0b6585bd44fe86344408d9d3ab8d3c17e254738a – wasn't in master until 3.4.0 was released the other week though.

PR #1725, opened 2016-10-12
Update package.json

jackycute:
fix engine version which could make more compatibility to new node version

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "url": "git://github.com/hakimel/reveal.js.git"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -19,7 +19,7 @@
     "url": "git://github.com/hakimel/reveal.js.git"
   },
   "engines": {
-    "node": "~4.1.1"
+    "node": ">=4.1.1"
```
demoneaux: Can we make this `>=4.0.0`?

hakimel:
Fixed in #1760

PR #1738, opened 2016-10-30
Made Reveal object use its own API for some internal navigation calls…

andrewcmyers:
… so that

navigation can be overridden externally in a uniform way. This requires
building the Reveal object early and patching it later with references to
the right functions (aka "Landin's knot"). Even more extensibility could be achieved by converting
fully to an OO style.

This is not a complete job of converting to the style I'm suggesting; thus far I only
indirected through the Reveal object for the navigateX() calls. I'm creating a pull
request mainly as a suggestion of a way to proceed.

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -162,9 +162,6 @@ module.exports = grunt => {
 
 	});
 
-	grunt.loadNpmTasks('grunt-contrib-clean');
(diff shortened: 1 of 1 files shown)


B45HF45H:
files unchanged

Naamboon:
#AndroidDevChallenge.

WritersWritingWordsbeingcharacters:
Duplicate of #

Issue #1681, opened 2016-09-01
How can I reuse the 3D rotation


martinko2009:
In the older version, it is default to the 3D rotation.But now in the new version, it is default to slide.How can I reuse the 3D rotation.Thank you.

christinakayastha:
You can use `data-transition='cube'`

demoneaux:
Older version of Reveal (<=2.6.2) had the `concave` transition enabled by default.

PR #1739, opened 2016-10-30
Add the cube and page transitions to the demo.html

### LINKED ISSUE #1718: It is feature requests that add to new transition.
We are add to new transition to turn a page of a book.
It is most standard for jQuery UI Plugin .
Regard,

christinakayastha:
The cube and page transitions are available but not easy to see, adding it to the demo.html so that users can find it (:

Closes #1718, #1681


hakimel:
Thanks for the PR but those transitions are deprecated, please see my comment in #1774

christinakayastha:
Gochha thanks!

PR #1740, opened 2016-10-30
Hide slide number when in paper or pdf print view

### LINKED ISSUE #1647: slide-number should not be displayed when printing
This is one of the controls that should be hidden when printing: it renders as the exact same value on every page (the current slide that was being viewed when printing was initiated.)

Easily fixed by adding `.reveal .slide-number` to the [list of classes hidden in paper.css](https://github.com/hakimel/reveal.js/blob/de2ab07334766e6df2e3e69cd58a820e40d6349e/css/print/paper.css#L33).

christinakayastha:
The slide number should be hidden when in pdf or print views

Closes #1647

### DIFF
--- a/css/print/paper.css
+++ b/css/print/paper.css
@@ -38,7 +38,8 @@
 	.share-reveal,
 	.state-background,
 	.reveal .progress,
-	.reveal .backgrounds {
+	.reveal .backgrounds,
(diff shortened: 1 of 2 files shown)


hakimel:
Makes sense, thanks!

PR #1744, opened 2016-11-10
collapsible fragments

kenegozi:
see README update for usage

### DIFF
--- a/README.md
+++ b/README.md
@@ -699,6 +699,25 @@ The display order of fragments can be controlled using the ```data-fragment-inde
 </section>
 ```
 
+#### Collapsible fragments
+
+You can define a fragment to be collapsible - i.e. it would not take any room when not the active one.
+This is useful for showing a progression of fragments when each one relpaces the one before it.
+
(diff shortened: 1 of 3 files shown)


hakimel:
Thanks. In your example, could you just use one fragment for each word? 
```
<span class="fragment">This</span>
<span class="fragment">This</span>
<span class="fragment">This</span>
```

hakimel:
Thanks. We already have something similar to this in the `current-visible` fragment style. A `current-visible` fragment will appear when it is the current fragment, and disappear again as soon as you move to a another fragment.

What you are proposing seems like the same thing, but instead of just fading in and out it would collapse via display none. That said, what do you think about implementing this as "fragment current-display" (As in "only display when it is the current fragment")?

All the CSS this would take is:
```css
.fragment.current-display:not(.current-fragment) {
	display: none;
}
```

kenegozi:
A simpler CSS for sure! would it support various containers (p, div, pre, etc.?)

PR #1747, opened 2016-11-15
Defaults in README are not current

spencermathews:
The `transition` and `backgroundTransition` defaults presented in the README are not up to date with reveal.js


hakimel:
Good catch.

PR #1751, opened 2016-11-17
use HeadJS for script loading in math plugin

FND:
reveal.js depends on HeadJS anyway, so we might as well use that.

Note, however, that I haven't actually tested this, as I'm not currently using this plugin. It's just something I stumbled across while looking at various plugins for reference. I don't see a reason why this wouldn't work though (HeadJS uses the file extension to decide between CSS and JavaScript).


hakimel:
Thanks but since this is already working and not dependent on another script, I'd rather leave it that way for flexibility.

reveal.js can be used without head.js, and if you do, you can load the reveal.js plugins any way you like.

PR #1752, opened 2016-11-17
mousescroll to change slides

matasoy:
You should add mousescroll event to change slides

`

var mousewheelevt=(/Firefox/i.test(navigator.userAgent))? "DOMMouseScroll" : "mousewheel" //FF doesn't recognize mousewheel as of FF3.x 
			if (document.attachEvent) //if IE (and Opera depending on user setting)
				document.attachEvent("on"+mousewheelevt, function(e){				
					if(e.wheelDelta == 120) Reveal.left(); else Reveal.right();
				})
			else if (document.addEventListener) //WC3 browsers
				document.addEventListener(mousewheelevt, function(e){
					if(e.wheelDelta == 120) Reveal.left(); else Reveal.right();
				}, false);
`


hakimel:
We already support this via the `mouseWheel` config value, try setting `mouseWheel: true` when initializing.

PR #1754, opened 2016-11-21
Allow add class to slide background

oanhnn:
Allow add class to slide background by `data-background-class`. 
Eg: `data-background-class="blurred"` and in `css` file:
```
.blurred {
    -webkit-filter: blur(5px);
    -moz-filter: blur(5px);
    -o-filter: blur(5px);
    -ms-filter: blur(5px);
    filter: blur(5px);
}
```
Background is blurred 5px and it look like http://oanhnn.github.io/slides/2016-11-21-unit-test-with-laravel.html#/4

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -777,6 +777,7 @@
 
 		var data = {
 			background: slide.getAttribute( 'data-background' ),
+			backgroundClass: slide.getAttribute( 'data-background-class' ),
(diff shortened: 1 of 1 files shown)


hakimel:
This is actually already possible – any non-reveal.js class you add to a slide gets mirrored on the background element. So all you should need is `<section class="blurred"></section>`.

PR #1756, opened 2016-11-22
Implementing a better data-trim

jmpp:
Inserting code snippets can sometimes be painful for the developer's indentation of HTML, constrained by the `<pre>` tag to pull code to the left like this :

```html
  <body>
    <div class="reveal">
      <div class="slides">

          <section>
            <section>
              <pre><code class="js" data-trim>
const fs = require('fs');

fs.writeFile('filename.txt', 'Hello World!', (err) => {
  if (err) throw err;

  console.log('Le contenu du fichier a bien été modifié.');
});
              </code></pre>
            </section>
          </section>

      </div>
    </div>
  </body>
```
✖ This breaks the editor's code folding, and makes more difficult the reading of the code when working on large presentations.

Now with the new smart `data-trim` attribute, each code snippet will be processed line-by-line to be automatically reindented to the left :

image

No more pain for the developer to maintain his presentation code indentation, can allow code folding in the editor like you would do in a normal situation :

```html
          <section>
            <section>
              <pre><code class="js" data-trim>
                const fs = require('fs');

                fs.writeFile('filename.txt', 'Hello World!', (err) => {
                  if (err) throw err;

                  console.log('Le contenu du fichier a bien été modifié.');
                });
              </code></pre>
            </section>
          </section>
```
✔

### DIFF
--- a/README.md
+++ b/README.md
@@ -301,6 +301,20 @@ Reveal.initialize({
 });
 ```
 
+If you wish to disable this behavior and do your own scaling (e.g. using media queries), try these settings:
+
+```javascript
+Reveal.initialize({
+
+	...
+
+	width: "100%",
+	height: "100%",
+	margin: 0,
+	minScale: 1,
+	maxScale: 1
+});
+```
 
 ### Dependencies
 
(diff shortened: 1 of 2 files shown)


jmpp:
@hakimel I read the CONTRIBUTING.md, please let me know if there is any oversight :)
Best.

hakimel:
This is great. Thanks a lot!

PR #1743, opened 2016-11-09
Bump node engine in packages.json

haf:

Issue #1729, opened 2016-10-17
Update node version in package.json


kolis:
Reveal.js does not install with the `yarn` package manager and node engine v6. I'm getting this error:
`error reveal.js@3.3.0: The engine "node" is incompatible with this module. Expected version "~4.1.1".`

I can install it using npm, and it seems to work, so I guess it is only the entry in `package.json` that needs updating.

kolis:
```
$ node --version
v6.8.1
```

msegado:
I also ran into this.
(@kolis, as a temporary workaround, yarn has a `--ignore-engines` flag you can use)

hakimel:
Updated in #1743

mrsum:
Hm... Still have troubles with reveal 3.3.0 and node v6.9.1

pirelenito:
Was this release? I just got this issue while trying to install it with `yarn add reveal.js`.

Btw, thanks for this awesome project! Really, it is awesome! ❤️

hakimel:
Just released this as part 3.4.0 🚀

PR #1760, opened 2016-12-06
Reduce required Node.js version to 4.0.0.

demoneaux:
The development workflow can be used with Node 4+, not just Node 6. We should avoid showing warnings and errors when users are installing this package via npm/yarn.

Ref. #1729, #1743.
Closes #1725.

### DIFF
--- a/README.md
+++ b/README.md
@@ -929,7 +929,7 @@ Reveal.initialize({
 
 Then:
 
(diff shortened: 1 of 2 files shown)


hakimel:
Thanks Ben.

hakimel:
This commit was added to master as well.

PR #1765, opened 2016-12-08
Enable passing options for parsing Markdown.

### LINKED ISSUE #1163: Enable SmartyPants option of Marked when rendering Markdown

demoneaux:
Closes #1163, #1495.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,18 @@ Special syntax (in html comment) is available for adding attributes to the slide
 </section>
 ```
 
+#### Configuring `marked`
+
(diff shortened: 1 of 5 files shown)


hakimel:
Great stuff! 👍

PR #1771, opened 2016-12-12
Resume video/audio playback if stopped by Reveal.js

### LINKED ISSUE #1770: Resume audio / video playback on 'present' slides.
Hi,

I'm using 3rd party video player in my slides, not 'data-background-video'. I was confused when video playback didn't continue when slide was shown again.

Ok, I debugged both Flowplayer and Reveal.js code, first I thought that Reveal was not doing it, because I was not using 'data-background-video'. After a 2 hours of debugging + POC:ing I found that REVEAL is pausing all video/audio playbacks on line [3128](https://github.com/hakimel/reveal.js/blob/master/js/reveal.js#L3128)

I also noticed that there is attribute 'data-ignore', BUT I CANNOT ADD THAT FOR VIDEO elements (because Flowplayer handles those) and yes, I would prefer that audio/video playback is stopped for slides which are not shown.

**Proper fix, is that Reveal.js can also resume playback for the video / audio elements which it did pause!**

I'm also using Flowplayer background video plugin, which removes controls from UI, so it cannot be resumed by user. My [demo](https://jsfiddle.net/RopoMen/0kx43k7u/)

BR,
Markku Roponen
PS. hoping to see fix soon ;)

RopoMen:
Fixes #1770 

Any comments? :)

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -3016,7 +3016,8 @@
 
 			// HTML5 media elements
 			toArray( slide.querySelectorAll( 'video, audio' ) ).forEach( function( el ) {
-				if( el.hasAttribute( 'data-autoplay' ) && typeof el.play === 'function' ) {
+				if( ( el.hasAttribute( 'data-autoplay' ) || el.hasAttribute( 'data-paused-by-reveal' ) ) && typeof el.play === 'function' ) {
+					el.removeAttribute('data-paused-by-reveal');
 					el.play();
 				}
 			} );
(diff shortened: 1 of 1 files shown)


RopoMen:
Adding 'data-paused-by-reveal' only if video / audio is not already paused. Now Reveal is not continuing playback for the media elements which it didn't pause.

RopoMen:
ping @hakimel

hakimel:
Good catch. That's inconsistent that we were pausing all videos, but only resuming if `data-autoplay` was present.

Has been merged into the dev branch.

Issue #1758, opened 2016-11-26
Alt-click element zooming broken with non-default width/height values


tferrin:
Setting the presentation resolution to values other than the default width: 960 and height: 700 causes alt-click element zooming to zoom to the wrong position on the slide.

To reproduce this problem set width and height in Reveal.initialize() to anything other than the defaults.  I create most of my presentations with width: 1280 and height: 1024, but setting any values other than 960/700 causes the problem.  Then view any presentation with multiple html elements on a slide and alt-click on one of the elements.  The slide is zoomed, but not focused on the chosen element.

Not sure if the new width and height are not getting passed to zoom correctly, or if the algorithm in zoom.js has a bug.

Tested with Firefox 50 and Google Chrome 54.

PR #1772, opened 2016-12-12
Fix for broken alt-click element zoom, ticket #1758.

tferrin:
Don't bother to zoom based on coordinates (left, top, width height), rather just zoom on a HTML element.


hakimel:
Thanks for the PR but we need to manually factor in reveal.js' current scale for zooming to work in all circumstances.

tferrin:
It's been a while since I did the PR, but at the time it did work for 
all circumstances, not just the one I originally opened the ticket for. 
I tested the change against general case of scaling to chosen screen 
size as well as the element scaling issue I originally noted.  I'd be 
happy to test again other cases too if I've missed something.

hakimel:
Gotcha. It's possible something might have changed since this was opened. I tried out using the element as a target earlier today but it the zoom area was offset incorrectly. I'm using a large display, so reveal.js is at a 1.5-2x scale. Feel free to give it another try if you want.

PR #1774, opened 2016-12-14
Add notes about the cube/page transitions.

demoneaux:
Closes #1739.
Ref. #1681, #1718.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,7 +248,7 @@ Reveal.initialize({
 	previewLinks: false,
 
 	// Transition style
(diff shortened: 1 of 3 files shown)


hakimel:
I actually intentionally removed cube/page from all comments and docs some time ago to deprecate them without breaking existing presentations. I think it's inconsistent how those two transitions make visual changes to the slide (semi-transparent black background and shadow), making them difficult to use in combination with other transitions or with mixed in/out transitions.

Will add a note about them being deprecated in the CSS to avoid confusion.

PR #1778, opened 2016-12-22
Add support for background-video for iPad by introducing controls on all mobile devices.

mbjedk:


hakimel:
Spent some time testing this but from what I can tell, the presence of `controls` doesn't have an impact on whether or not videos play on iPad.

Videos definitely weren't playing, but I found another solution for that in 7daa0e04de7d0ab2bae3159faae0b18d4eb3c8d3. I've updated the online demo, feel free to test by going to this link and stepping downwards to the slide with a video background: http://lab.hakim.se/reveal-js/#/10

PR #1780, opened 2016-12-23
support complex markdown

ccqgithub:
support <textarea data-template> `some code` </textarea>

### DIFF
--- a/plugin/markdown/markdown.js
+++ b/plugin/markdown/markdown.js
@@ -42,8 +42,8 @@
 	 * element. Normalizes leading tabs/whitespace.
 	 */
 	function getMarkdownFromSlide( section ) {
-
-		var template = section.querySelector( 'script' );
+		// support <textarea data-template> `some code` </textarea>
+		var template = section.querySelector( '[data-tempate]' ) || section.querySelector( 'script' );
 
 		// strip leading whitespace so it isn't evaluated as code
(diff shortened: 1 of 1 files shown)


demoneaux:
What's the use case for this? Including Markdown within HTML tags will cause it to be parsed by the browser first, hence the resulting Markdown might not be suitable for rendering.

ccqgithub:
@demoneaux 
the the html tag maybe appear with the code snippet。
 
```
<script>
// the script expamle in mardown code 
</script>
```

demoneaux:
I believe that escaping the `</script>` closing tag is enough, using something like this should work:

```html
<section>
<script type="text/template">
# Heading

    <script><\/script>
</script>
```

ccqgithub:
that will make highlight troubles, i know escaping can do that, but i will escape all code in the code of md ?
```
<section>
<script type="text/template">
   # Heading
   ` ` `
       <!DOCTYPE html>
    <html lang="en">
    <head>
      <meta charset="UTF-8">
      <title>Document</title>
    </head>
    <body>
      <script></script>
    </body>
    </html>
   ` ` `
</script>
```

hakimel:
Seems like this makes sense. Using a textarea wrapper avoids the conflict with `<script>` without any drawback that I know of.

The commit has been merged into dev, and the readme was updated to recommend using this over `<script type="text/template">`.

PR #1782, opened 2016-12-28
width and height configuration via callback

PascalPflaum:


hakimel:
Thanks for the PR but I don't think it makes sense to set those values via callback. We already have a `Reveal.configure()` API method that lets you change any config value after initialization. If you have a dynamically calculated width/height you could do something like:

```javascript
function getWidth() {
  Math.random() * 1000;
}

Reveal.initialize({ width: getWidth() });

window.addEventListener( 'resize', function() {
  Reveal.configure({ width: getWidth() });
} );
```

PR #1787, opened 2017-01-09
Add config to enable/disable auto-slide controls

abhijitvalluri:
Add the ability to control whether the auto-slide configuration is displayed using the config.

Signed-off-by: Abhijit Kiran Valluri <avalluri2@bloomberg.net>

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -107,6 +107,9 @@
 			// by using a data-autoslide attribute on your slides
 			autoSlide: 0,
 
+			// Enable slide controls animated button
+			enableSlideControls: true,
+
 			// Stop auto-sliding after user input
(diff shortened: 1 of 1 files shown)


hakimel:
Hey! If you set `autoSlideStoppable` to false, that will hide the controls. The auto-sliding can't be stopped then, but if there's a separate option for only hiding the controls and you do stop the progression there would be no way to resume so that doesn't seem ideal.

Does `autoSlideStoppable` solve what you were trying to accomplish or is there something I'm missing?

abhijitvalluri:
Yes, the `autoSlideStoppable` does what I need. I guess I don't need this potentially conflicting setting anymore.

PR #1794, opened 2017-01-15
Remove 'history: true' from index.html

pkubowicz:
Readme says the default is 'history: false', so it may be surprising
for users - it's easy to miss the contradicting setting in index.html


hakimel:
I agree this was confusing. Thanks!

PR #1800, opened 2017-01-20
Allow showNotes to be triggered by query parameter

amb-it-ion:
My regular use case for showNotes is for PDF print-outs. Restricting the showNotes setup to configuration only requires me to change a file. Allowing to add "show-notes" to the query parameters makes life easier again.


hakimel:
Thanks for the PR but this is already be possible via `?showNotes=true`, for example: http://lab.hakim.se/reveal-js/?print-pdf&showNotes=true (first speaker notes are on the 10th slide, titled "Fragments").

On a related note I recently made it possible to print them on a separate page by setting hte value of showNotes to 'separate-page': http://lab.hakim.se/reveal-js/?print-pdf&showNotes=separate-page

Good idea to mention this in the docs where PDF printing is covered. I'll add something.

PR #1803, opened 2017-01-24
correct grunt option usage

bitjson:
The command in the readme fails using the latest version of Grunt. This fixes the issue.


hakimel:
thanks!

PR #1816, opened 2017-02-05
added toggleHelp function

Frazer:
this way a key can be given ability to toggleHelp on and off.  Previously Reveal.showHelp could open the help screen, but no way to close it.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -1629,6 +1629,18 @@
 	}
 
 	/**
+	 * Open or close help overlay window.
+	 */
+	function toggleHelp(){
+		if( dom.overlay ) {
+			closeOverlay();
+		}
+		else {
+			showHelp( true );
+		}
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks for the PR! I agree, `toggleHelp` is more in line with our other API methods.

All other "toggle" methods support an `override` argument, which essentially lets you force it to either toggle on or off. Would you mind adding a flag like that for the toggleHelp method too? You can see how it's implemented for `toggleOverview` here https://github.com/hakimel/reveal.js/blob/master/js/reveal.js#L2054-L2070

Also please submit all pull requests towards the dev branch. That's where all the work happens between releases.

hakimel:
Thanks for the updates, @Frazer! 

I added `toggleHelp` to the docs and removed the `showHelp` API method since it's redundant.

PR #1818, opened 2017-02-06
removed double element.webkitRequestFullscreen || in enter full screen

Frazer:


hakimel:
Thanks but those two are actually not identical – note how one uses lowercase "s" in "Screen" and the other uses uppercase "S". They're both needed to cover all versions of webkit.

PR #1823, opened 2017-02-07
Update to README.md for PDF printing

gilxa1226:
I found that when trying to print to PDF, there was a block of code missing from the HTML, this block is necessary for the PDF layout to be rendered properly and I didn't see it anywhere in the README so I added it in.


hakimel:
Thanks! Merged with a couple of tweaks: https://github.com/hakimel/reveal.js/blob/master/README.md#pdf-export

PR #1826, opened 2017-02-11
Slide-thumbnails plugin

burnpanck:
The main use of this is to enable a thumbnail based slides overview, in the hope of getting better performance. The corresponding overview mode can be found in another [branch of mine](https://github.com/burnpanck/reveal.js/tree/thumnail-based-overview).

This plugin generates a thumbnail of the initial fragment state of each slide. Arguably, the final fragment state would be more appropriate. It simply waits 1.5s per slide for the transitions to pass, but does not wait e.g. for MathJax to finish processing. Support from the presentation side could help here.


gelinger777:
How to get image of thumbnail as base64??

Issue #1155, opened 2015-03-05
Scrolling in overview mode on iPad not fluid


Pirouette04:
Hi,

I'm thinking of use reveal.js in a projet, at this time I see just one stopper for this. The fluidity of the overview when I'm scrolling. 

Is the actual fluidity in overview mode is normal? 

Thanks!

hakimel:
The overview isn't very smooth in its transitions on some devices. It's just a very heavy view to render since it shows so many slides at once, transformed down to a smaller size.

I haven't found a way to make it any faster but if it's unbearably slow I may revert back to not having any transition at all instead.

PR #1827, opened 2017-02-11
WIP: Thumbnail based overview

### LINKED ISSUE #94: Overview mode is laggy
When I open overview mode with a slideshow that has about 20 pictures, it lags and glitches. Suggested fix: changing images so that they resize to a smaller size immediately before overview mode is opened.

I might submit a pull request to fix this sometime in the next few days if someone else doesn't.

burnpanck:
This is an attempt to fix #94 and #1155, by replacing the actual slides with thumbnails.
However, it makes use of additional CSS wrappers to hide the actual slide content,
introduced in PR #1009, which will not get merged.
Possibly, there is another way to hide the slide contents but still enabling
the thumbnail to be placed inside the slide `section`.
However, for me, separating slide placement in the browser view
and content layout in the slide is a good thing, so I stick with that approach.

It looks for thumbnails in "slide-thumbs/slide-%h-%v.png",
here %h and %v refer to the slide index.
This goes together with my plugin "slide-thumbnails".
Note that, currently, no detection on availability of thumbs is done.
If they are missing, you get a blank overview.


hakimel:
I don't think we want to switch to an image based thumbnail overview since you'd have to go through the extra manual step of generating images first.

Sorry about the late reply on this, slowly trying to catch up to PRs.

PR #1831, opened 2017-02-17
Correctly strip leading white-space from markdown

jerrykan:
If the markdown contains something that is indented by more that the `leadingTabs`/`leadingWs` then extra white space is incorrectly removed. ie the following example:

```
    <section data-markdown>
    some text
       indented text
          more indented text
    </section>
```

would result in the following markdown:

```
some text
   indented text
  more indented text
```

We can work around this problem by using a function to generate the
replace value.


hakimel:
This commit was merged via #3358

PR #1833, opened 2017-02-17
allow configuration of which view slide number appears on - fixes #1791

### LINKED ISSUE #1791: Include slide number in speaker notes view
It would be very useful to display the current slide number and the total number of slides in the speaker notes view, along with the time

malcomio:


hakimel:
Thanks! Merged with two changes;
- Switched the value from 'notes' to 'speaker', so it reads more like "show the slide number to the speaker".
- Added a third value 'print' which can be used to only show notes when printing to PDF.

PR #1836, opened 2017-02-24
Allow whitespace on background-images, w3 compliance

RobertBaron:
https://www.w3.org/TR/CSS2/syndata.html#value-def-uri

Some characters appearing in an unquoted URI, such as parentheses, white space characters, single quotes (') and double quotes ("), must be escaped with a backslash so that the resulting URI value is a URI token: '\(', '\)'.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -3085,7 +3085,7 @@
 
 				// Images
 				if( backgroundImage ) {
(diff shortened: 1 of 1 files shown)


bnjmnt4n:
Merging as this seems to be the right behaviour to generate valid URLs for our generated CSS.

Issue #1835, opened 2017-02-22
Disable pointer in full-screen mode?


emveeoh:
Is there any way to disable the pointer when the browser is full-screen?

PR #1837, opened 2017-02-25
hide mouse if inactive after 5 seconds #1835

romdum:


romdum:
It's done ! :)

gabyx:
Can we please merge this feature :-)?

hakimel:
I like the feature and was missing this myself the other day. This option is now available in the dev branch now and defaults to hiding after five seconds.

Wanted to change a couple of different things to make this work when reconfiguring reveal.js and touch the DOM a little bit less so I ended up writing a different implementation in 7b707696b40a98e717d91e24162869f6d9c22957

PR #1841, opened 2017-03-11
remove up and down arrows when disabled

### LINKED ISSUE #1839: Show only left/right control arrows
Is there an easy way to show only horizontal navigation arrows and not the up/down buttons? For a strictly horizontal presentation it might be confusing to show vertical controls that don't do anything.

mvpowers:
This fixes #1839 by hiding the up and down arrow buttons when there is not a slide to move to.

### DIFF
--- a/css/reveal.css
+++ b/css/reveal.css
@@ -243,6 +243,9 @@ body {
   border-bottom-width: 22px;
   border-bottom-color: #000; }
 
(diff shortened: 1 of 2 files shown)


hakimel:
The new controls (reveal.js 3.6.0) automatically hide the up/down arrows until you get to a vertical slide.

PR #1843, opened 2017-03-14
print-pdf using callback iso timer

riezebosch:
Use window.callPhantom icw page. onCallback to wait for pdf-ready event.
From: http://stackoverflow.com/a/28925479/129269


riezebosch:
It was working fine on current release version of reveal.js but on master it tends to be too fast because I'm missing syntax highlighting in my output... Stay tuned.

riezebosch:
updated commit to dev-branch iso master

hakimel:
Nice! Didn't know about callPhantom/onCallback.

PR #1844, opened 2017-03-14
print-pdf layout slide contents to fit stretch elements

riezebosch:
layout() and thereby layoutSlideContents() is never invoked
when isPrintingPdf() is true so stretch elements are not
sized correctly.
This is ensured now by invoking the layoutSlideContents
from SetupPDF(). There seems no need to scale down
like in other PR's is suggested.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -614,6 +614,9 @@
 		document.body.style.width = pageWidth + 'px';
 		document.body.style.height = pageHeight + 'px';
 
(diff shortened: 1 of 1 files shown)


riezebosch:
Same as https://github.com/hakimel/reveal.js/pull/1416 but without the scaling. For me it is working perfectly on a rather complicated (=huge) slide deck.

hakimel:
Tried a few examples and this seems to be working well. Thanks!

PR #1849, opened 2017-03-17
Optimizations around notes plugin

danbroooks:
This commit makes some tweaks around the `handleStateMessage` function inside the notes plugin's `notes.html`.

For larger presentations the notes plugin is running 3 different presentations at once (presentation, upcoming slide, preview slide), and this can slow the performance of the main presentation. A potential fix is to slightly delay the update of the notes preview windows, allowing the presentation to update first, then the notes plugin to follow shortly after.

A slightly longer delay has been added for the `notes-only` layout as this layout isn't showing any of the previews at all, yet is still running them and updating them in the background.

### DIFF
--- a/plugin/notes/notes.html
+++ b/plugin/notes/notes.html
@@ -363,43 +363,40 @@
 				/**
 				 * Called when the main window sends an updated state.
 				 */
-				function handleStateMessage( data ) {
+				var handleStateMessage = (function ( currentTimeout ) {
+					return function ( data ) {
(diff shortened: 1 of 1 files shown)


danbroooks:
I've come to notice that this gets the presentation stuck into a loop if slides are changed too quickly, I guess the socket events being fired about end up getting the presentation stuck... instead of closing this is there anything else I could do to achieve a similar result?

PR #1851, opened 2017-03-18
Fix data-background-video autoplay on iOS Safari

astone123:
Add 'autoplay' and 'playsinline' attributes to data-background-video elements so that they autoplay on iOS Safari browser


hakimel:
🙌

Issue #1857, opened 2017-03-27
gruntfile.js don't work with old version of node


HaoranYi:
The following is the error message when running an old version of node.

D:\workshop\reveal.js>npm start

> reveal.js@3.4.1 start D:\workshop\reveal.js
> grunt serve

D:\workshop\reveal.js\Gruntfile.js:133
                                files: root.map(path => path + '/*.html')
                                                      ^
Loading "Gruntfile.js" tasks...ERROR

A fix could be like this. 

```
diff --git a/Gruntfile.js b/Gruntfile.js
index 96a4f52..11486f8 100644
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -130,10 +130,10 @@ module.exports = function(grunt) {
                                tasks: 'css-core'
                        },
                        html: {
-                               files: root.map(path => path + '/*.html')
+                files: root.map(function(path) { return path + '/*.html';} )
                        },
                        markdown: {
-                               files: root.map(path => path + '/*.md')
+                files: root.map(function(path) { return path + '/*.md'; })
                        },
                        options: {
                                livereload: true
```

demoneaux:
Using a recent version of Node is recommended as Node 0.10 and 0.12 are already past their end of life. We are also in the process of upgrading our grunt dependencies (#1884), which will require Node v4+ for one of the libraries. Hence I'll be closing this issue.

PR #1858, opened 2017-03-27
make gruntfile.js work with old version of node

HaoranYi:
This is the pull request for issue 1857, which make gruntfile.js works with old version of node.js. 
Can someone take a look? Thanks!


demoneaux:
See https://github.com/hakimel/reveal.js/issues/1857#issuecomment-298299190.

PR #1864, opened 2017-04-04
Change data background hash generation

aloupfor:
I saw an unexpected behavior in the data-background-hash generation.
I don't think that it is intended to have a multiple concatenation of "null".

In the demo.html, there is:
```
<section data-background="#dddddd">
```
which gives
```
data-background-hash="#ddddddnullnullnullnullnullnullnullnull"
```

I am a JavaScript beginner.
I saw that a way to prevent this is to use the `filter` method with the `Boolean` constructor.
I don't know which browsers are currently supported but `filter` is only available from IE9.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -882,16 +882,18 @@
 		// Create a hash for this combination of background settings.
 		// This is used to determine when two slide backgrounds are
 		// the same.
-		if( data.background || data.backgroundColor || data.backgroundImage || data.backgroundVideo || data.backgroundIframe ) {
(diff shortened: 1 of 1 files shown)


demoneaux:
Hi @aloupfor! The use of `null` in data-background-hash allows all the options to be included when comparing between hashes. Although unlikely, there might be cases where concatenation without the null would lead to the same property even when using different options. The use of null will allow us to ensure that this is easily differentiated and unlikely to happen.

aloupfor:
Thank you for reviewing this PR and for your explanation.

PR #1866, opened 2017-04-05
replace cdn.mathjax.org with cdnjs

christianp:
cdn.mathjax.org is shutting down: https://www.mathjax.org/cdn-shutting-down/

This changes the default MathJax URL in the math plugin, as well as
references in README.md and test/examples/math.html


hakimel:
Had no idea that was happening – thanks for the fix.

Issue #1874, opened 2017-04-22
Reducing size of `pre` block around `code` block


draptik:
Currently the `drop-shadow` border around `pre` blocks takes up unneeded vertical space around nested `code` blocks. Is this the intended behaviour?
Here is an example:
```xml
<section>
	<pre>
		<code data-noescape data-trim class="lang-csharp hljs">
			public interface IFoo
			{
				void DoMagic(string s);
			}
		</code>
	</pre>
</section>
```

![screenshot from 2017-04-22 15-35-57](https://cloud.githubusercontent.com/assets/313058/25304947/0666fe94-2772-11e7-95e9-f7a290e8a6ad.png)
(Notice the space above and below the code block and the border)

Am I doing something wrong here?

Replacing
```css
reveal pre {
  display: block;
```
with
```css
reveal pre {
  display: inline-grid;
```
in `css/theme/template/theme.scss` (and running `grunt css-themes`) seems to fix the issue for me:
![screenshot from 2017-04-22 15-36-54](https://cloud.githubusercontent.com/assets/313058/25304995/1e7fadb8-2773-11e7-91ca-faa66d97549c.png)

demoneaux:
You're leaving a line of whitespace between the `<pre>` and `<code>` tags. The default Reveal styles only the `code` tag, hence there appears to be this issue for you. While I'm not too sure what the solution should be, I don't believe modifying the display is the right way to go as it can lead to unexpected behaviour. For now, a quick fix for you is to adjust the whitespace between pre and code.

draptik:
So removing the linebreak and writing
```xml
<pre><code>
...
</code><pre>
```
instead of
```xml
<pre>
  <code>
  ...
  </code>
<pre>
```
fixes the problem.

@demoneaux Thanks!

merlinpatt:
Is there a fix to do this without removing the line break between pre and code tags?

PR #1875, opened 2017-04-22
pre code layout change

draptik:
- match size of pre block around code block
- see https://github.com/hakimel/reveal.js/issues/1874 for details


draptik:
PR is not needed. So correct answer in https://github.com/hakimel/reveal.js/issues/1874

Issue #1630, opened 2016-06-23
Keyboard Bindings Trigger when using electron <webview> tag


carlskii:
Hi,

I've embedded a presentation into an electron app and I'm using the <webview> tag since it has more controls over <iframe>. However, when typing in within the <webview> the keyboard navigation shortcuts are being fired. This doesn't happen with <iframe> 

Can this be fixed or can I disable the keyboard shortcuts completely ?

carlskii:
So I tried disabling the keys using : 

```
Reveal.configure({
                keyboard: {
                    32: null, // don't do anything when SPACE is pressed (i.e. disable a reveal.js default binding)
                    83: null,
                    66: null,
                    82: null,
                    79: null
                }
            });
```

However the 83 (s) code gets ignored and opens the presenter notes when pressed if within the  <webview>

PR #1885, opened 2017-04-30
Key Binding API

### LINKED ISSUE #1832: Keyboard binding `s` does not override Speaker Notes
I have: 

```
      Reveal.configure({
        keyboard: {
          68: 'next', //d
          65: 'prev', //a
          88: 'down', //x
          87: 'up', //w
          83: function(){location.reload(); return false; }//s
        }
      });
```

This piece of code reloads my page on `s` pressed, but also opens speaker notes. How can I disable it?

denehyg:
Adds a Key Binding API that allows plugins to programmatically add and remove key bindings to Reveal's key handling.
- plugins can make use of Reveal's pre-processing logic for key handling so they don't have to replicate (for example, ignoring key presses when paused)
- (optionally) includes key in the help overlay. Unlike Reveal.registerKeyboardShortcut() when key bindings are removed they are also removed from the help overlay
- reduces the need for plugins to directly add `keydown` event listeners that can't be overridden by user defined bindings using `keyboard` config option
- incorporates API usage into notes plugin (should fix #1832 and #1630)
- passes `keydown` event to config.keyboardCondition() to allow plugins to filter keyboard based on keycodes. Use case: the reveal.js-menu plugin blocks Reveal keyboard interaction (using keyboardCondition) when the menu is open but using the key binding API to toggle the menu fails to close the menu unless the 'm' key is allowed to pass through to Reveal.


denehyg:
This will also provide access for plugins to respond to `Reveal.triggerKey()`, which is currently not possible as `triggerKey()` directly invokes `onDocumentKeyDown()` without generating key press.

hakimel:
This is a great addition. Thanks for a well structured PR with docs, comments and all 🙌

denehyg:
You’re welcome, glad it was useful.

PR #1887, opened 2017-05-03
adds `useAvailablePort` option when serving page with connect.

tristansokol:
I frequently have to specify a port because `8000` is taken. This feature should help out with that.


hakimel:
Very useful, thanks!

PR #1893, opened 2017-05-05
Update highlight.js to 9.11.0

mabhub:


hakimel:
Thanks but looks like it was already updated to 9.11.0 in another PR.

PR #1898, opened 2017-05-11
Fix typo

bentsai:
"proceding" → "proceeding"


bnjmntn:
@hakimel should we merge this?

PR #1909, opened 2017-05-21
Better Search and Modified Zoom

linux-man:
Open/close with Control+Shift+f
Find words inside SPAN
Don't duplicate findings

Block dimension calculation is not accurate, so zoom is made on mouse coordinates
Change default modifier to ctrl - alt don't work on Linux
New parameter zoomLevel: default 2

Also, can you bless (or curse) the [reveal.js editor](https://github.com/linux-man/unreel) I'm writing?

### DIFF
--- a/demo.html
+++ b/demo.html
@@ -399,6 +399,7 @@ Reveal.addEventListener( 'customevent', function() {
 					{ src: 'plugin/markdown/marked.js', condition: function() { return !!document.querySelector( '[data-markdown]' ); } },
 					{ src: 'plugin/markdown/markdown.js', condition: function() { return !!document.querySelector( '[data-markdown]' ); } },
 					{ src: 'plugin/highlight/highlight.js', async: true, callback: function() { hljs.initHighlightingOnLoad(); } },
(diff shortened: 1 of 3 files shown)


bnjmntn:
@hakimel what do you think?

hakimel:
Search improvements look great.

I don't want to remove the zooming based on element bounds, if it's not working correctly we should fix it instead. It's useful to be able to zoom specific elements rather than a point when you want to, for example, enlarge a whole code block.

How about adding the point zoom using a different shortcut than alt + click?

linux-man:
I turned to point-zoom because I couldn't fix the element zoom. The biggest problem is on fullscreen. Even when it's good on a maximized window, turning fullscreen send the elements to strange places. Maybe the issue is browser-dependent.

hakimel:
This has been merged into the dev branch 🙌

Played around with the point based zoom and I like how it's more predictable than element based zooming, since element bounds aren't always what you expect them to be.

ctrl + click brings up the context menu in OS X so I've made the shortcut platform dependent (ctrl for Linux, alt for all others).

Issue #1896, opened 2017-05-09
Client does not follow master on URL open


OleMussmann:
I would like to jump to specific slides on the master presentation with URLs on RFID tags. The master-client setup works well for slide changes. It does, however, not work for opening URLs. Once I manually change slides on the master, the client 'locks' on again, and follows. Is there any way to make the client follow on opening URLs?

OleMussmann:
Fixed in [pull request 1917](https://github.com/hakimel/reveal.js/pull/1917).

bnjmntn:
Closed by #1917.

PR #1917, opened 2017-06-06
Allow multiplexing on URL load

OleMussmann:
This fixes [issue 1896](https://github.com/hakimel/reveal.js/issues/1896).

The multiplexing now also works on opening URLs. This allows the links to be used as a remote control to jump to specific slides.

Are there any concerns with this approach?

### DIFF
--- a/plugin/multiplex/master.js
+++ b/plugin/multiplex/master.js
@@ -19,6 +19,9 @@
 
 	};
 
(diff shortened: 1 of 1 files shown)


bnjmntn:
@hakimel what do you think?

hakimel:
Tested and this seems to be working well. Thanks @OleMussmann!

PR #1923, opened 2017-06-18
allow specifying MathJax config with ?config=

dotlambda:
I want to specify MathJax configuration in the url.
This feature would be nice to have for using reveal.js with pandoc.


bnjmnt4n:
This request is too specific for most users, and seems like something that can be addressed by customizing your own `Reveal.initialize` call by checking the query string first.

bnjmnt4n:
Sorry, I misread the PR. This seems feasible.

bnjmnt4n:
Actually looking at it again, this seems to generate edge cases where the user might expect both configs in the URL and in the options to work, which is not intuitive. Since there is already an option to configure this, I don't think this is necessary.

Issue #1672, opened 2016-08-22
iframes are not preloaded according to viewDistance


mispy:
Adding iframe[data-src] to the selector here makes it behave as expected: https://github.com/hakimel/reveal.js/blob/bac187f3a0f7e8f92d33ca0258fe2e96553d2930/js/reveal.js#L2887

Is this omission intentional? If so, it should probably be documented in the README. :)

kp77:
Are there any plans for fixing this one?

gbrowdy:
simply adding `iframe[data-src]` to the selector array in the line of code referenced above causes iframe lazy loading to work for me.  I'll send a PR for this.

hakimel:
This is intentional. I've added a note to the readme to explain the rationale: https://github.com/hakimel/reveal.js/tree/dev#lazy-loading-iframes

PR #1938, opened 2017-07-05
added detection of iframe[data-src] to showSlide 

gbrowdy:
To allow for lazy loading of iframes
issue #1672

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -3074,7 +3074,7 @@
 		slide.style.display = config.display;
 
 		// Media elements with data-src attributes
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks for the PR but this behavior is intentional. I added a note to the readme that describes why: https://github.com/hakimel/reveal.js/tree/dev#lazy-loading-iframes

Might be worth making this configurable in the future though, so that you can change it depending on what type of content you want to show in your frames. I tend to run a lot of heavy JavaScript experiments in my talks and don't want them to load before becoming visible since they make my computer fans go 🔥🔥🔥

PR #1943, opened 2017-07-23
Add capabilities to run revealjs in docker container

jsenon:
hi,

I propose to add capability to launch reveal.js on docker container

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM node:latest
+
+RUN apt-get update \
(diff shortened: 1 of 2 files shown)


riezebosch:
Why nodejs? It will run just fine on a static HTML server like nginx.

bnjmnt4n:
This seems too specific to be included in the Reveal.js core repository. Do feel free to upload your Dockerfile and create a separate repo if there are features you want to share with others.

PR #1950, opened 2017-07-28
Fix whitespace detection and stripping in getMarkdownFromSlide

puyo:
I had this code in my slide:

    ```elixir
    defmodule YouTube do
      def search_videos(query) do
        Repo.all(from video in Video,
        where: query in video.search_terms)
      end
    end
    ```

Before:

screen shot 2017-07-28 at 1 55 27 pm

After:

screen shot 2017-07-28 at 1 56 00 pm

Similar to https://github.com/hakimel/reveal.js/pull/1847 and https://github.com/hakimel/reveal.js/pull/1831


puyo:
Seems fixed now! Thank you!

Issue #1584, opened 2016-05-09
PDF export: Add option to create one PDF page per fragment


flying-sheep:
This can’t be done using just CSS.

We would need to enumerate the fragments for each slide and determine their order (indices) as normal.

but then, for the CSS export, we would need to do sth. like this:

``` js
const fragmentIndices =
    toArray(sortFragments(currentSlide.querySelectorAll('.fragment')))
    .map((f, i) => i)

let lastSlide = currentSlide
for (const i of fragmentIndices) {
    lastSlide = lastSlide.parentNode.insertBefore(
        lastSlide.cloneNode(true),
        lastSlide.nextSibling)

    const fragments = toArray(lastSlide.querySelectorAll('.fragment'))
    for (cost fragment of fragments.slice(0, i+1) {
        fragment.classList.add('.visible')
    }
}
```

falexwolf:
Yes, this would be very helpful to have.

ninjin:
It would indeed be a very appreciated feature for times when you are forced to export your slides to PDF when you are not given the option to present using your own computer and there is no Internet connection, as is the case for me personally tomorrow.  There is a [hack posted on the issue tracker](https://github.com/hakimel/reveal.js/issues/808#issuecomment-127222420) to get around this current limitation, but it is far from optimal as it exports every slide as an image.

ThomasWeinert:
I build it into a plugin. https://github.com/ThomasWeinert/reveal-print

bnjmnt4n:
This seems like a pretty good idea, although having to duplicate nodes might cause additional side effects. I think this should belong in a plugin for now: any other opinions?

hakimel:
This was added in #1955 which has been merged into the dev branch.

I like this as the default behavior since it better mimics how fragments work. That said, there's a config option for disabling it if you want the old behavior.

commonsguy:
@hakimel I do not see such a config option (export PDFs with one page per slide, not one page per fragment) documented in [the configuration section](https://github.com/hakimel/reveal.js#configuration) or [the PDF export section](https://github.com/hakimel/reveal.js#pdf-export).

PR #1955, opened 2017-08-08
PDF export: add option to export one page per fragment

koehlma:
Allows to export one PDF page per fragment by appending `?pdf-print-fragments` to the URL.

Fix issue: #1584


hakimel:
This looks great–thanks!

I've merged your commits into the dev branch and made a few changes. Most importantly, this is now enabled by default when printing to PDF. To disable separate pages for fragments you'll need to set the `pdfSeparateFragments` config option to false.

Ir1d:
@hakimel Hi, is there a plan on when this feature will be released into npm? Thx!

PR #1958, opened 2017-08-12
Block F5 in speaker notes window, avoid disconnects

ThomasWeinert:
I noticed that because my presenter sends [F5] as presentation start key.


hakimel:
Good idea, thanks!

PR #1967, opened 2017-08-24
Changed a selector that was breaking print mode.

lanmaster53:
The old CSS selector failed to set the styled elements in print mode due to how the structure of the HTML changes. This problem appears to exist in other themes as well.


hakimel:
Good catch, thanks. This needed to be fixed in all themes, so I applied the fix to the theme template SCSS file and rebuilt all of the CSS in 0c946ae18b3773895bae259802108df11fade6d9

Issue #1970, opened 2017-08-27
Code appearing in headers is capitalized


ldionne:
I often want to write code in headers (usually h3 or h4's). Unfortunately, Reveal will capitalize the code, which is basically never what I want. Indeed, code should always be shown verbatim.

Here's a minimal reproduction:

```html
<!doctype html>
<html>
    <head>
        <link rel="stylesheet" href="css/reveal.css">
        <link rel="stylesheet" href="css/theme/league.css" id="theme">
    </head>

    <body>
        <div class="reveal">
            <div class="slides">
                <section>
                  <h3>normal text is capitalized, fine</h3>
                  <h3><code>SomeClass</code> is capitalized too, not fine</h3>
                </section>
            </div>
        </div>

        <script src="js/reveal.js"></script>
        <script>
            Reveal.initialize({});
        </script>
    </body>
</html>
```

To work around this, I add the following CSS:

```html
<style>
  .reveal h1 code, .reveal h2 code, .reveal h3 code, .reveal h4 code {
    text-transform: none;
  }
</style>
```

#### Before the CSS:
<img width="1029" alt="someclass - capitalized" src="https://user-images.githubusercontent.com/700834/29755028-d284c2b8-8b45-11e7-90a5-f122bc3e5575.png">

#### After the CSS:
<img width="1109" alt="someclass - not capitalized" src="https://user-images.githubusercontent.com/700834/29755030-d6c8c144-8b45-11e7-8679-3423425af19b.png">

This may not be the right fix (I don't know much CSS), but I think fixing this would be both worthwhile and easy.

sfairchild:
I just came across this while making a new presentation. After digging into what is happening it is the theme is adding `text-transform: uppercase`

I fixed this for myself by adding to my own stylesheet 
```
.reveal h1 span.no-transform,
.reveal h2 span.no-transform,
.reveal h3 span.no-transform,
.reveal h4 span.no-transform,
.reveal h5 span.no-transform,
.reveal h6 span.no-transform { text-transform: none; }
```

Then is each header where there is something I don't want uppercased I just add `<span class="no-transform">SomeClass</span>`

imho this is better than messing with the global style and breaking all other presentations

ldionne:
> imho this is better than messing with the global style and breaking all other presentations

1. Why would anyone want text inside `<code>` tags to be capitalized? I think the right thing is obviously not to capitalize it, would you disagree (ignoring backwards compatibility, see below)?
[...]

PR #1978, opened 2017-09-06
Prevent text-transform inside code tags

sfairchild:
This removes ```text-transform: uppercase;``` from within ```<code>``` even if they are used inside a header tag. This is in reference to issue #1970


hakimel:
Thanks! Your commit has been merged into the dev branch and will make its way over to master with the next release.

PR #1991, opened 2017-09-21
Backport console syntax in highlight.js

jreybert:
Based on highlight.js commit
https://github.com/isagalaev/highlight.js/commit/363e67e7196ff62eb5fcc3bcf3725e7389bdee13

I don't know what is the policy, if you backport some feature from external libs or if you fetch entire new versions.


bnjmnt4n:
Hi @jreybert! We update the entire version of highlight.js once in a while, so backports like these are not required. Thanks anyway!

PR #2009, opened 2017-10-24
font-variant vs text-transform

rphlypo:
Using the `small-caps` font-variant rather than the `uppercase` text-transform attribute for the headings.


hakimel:
Thanks for the suggestion but I prefer regular uppercase for the reveal.js themes.

Issue #1822, opened 2017-02-07
Code highlighting enters very slow loop on some C# templates


Siderite:
Simple code:
```
<pre><code class="hljs">
    private static void ensureContainer()
    {
        if (_container != null) return;
        lock(_lock)
        {
            if (_container != null) return;
            _container = new Container();
            _container.Register&lt;IWorldDominator, WorldDominator&gt;();
            _container.Register&lt;ITanks, Tanks&gt;();
            ...
        }
    }
</code></pre>
```
It makes the page load for a long time in order to highlight the code. If I remove the Register lines, the problem disappears.

demoneaux:
Not too sure what's causing this, will try to debug soon.

BDav24:
Upgrading highlight.js to v9.12.0 seems to fix the problem.

bnjmnt4n:
Closing as highlight has been updated to a more recent version and we can't duplicate this anymore.

PR #2016, opened 2017-10-27
Fix slow highlighting #1822

BDav24:
Fixes https://github.com/hakimel/reveal.js/issues/1822


bnjmnt4n:
We've already updated highlight to a more recent version, and we will continue to do so before new releases of Reveal.

PR #2025, opened 2017-10-31
add fragment: highlight-underline

ikarulus:
I added another highlight-fragment: underline

sdfsaf


bnjmnt4n:
@hakimel what do you think about adding another fragment?

hakimel:
Sure, I'm all for adding new fragment styles.

From what I can see the added test seems identical to the other fragment tests, with the exception of using a different fragment style. If that's the case can you please remove the test from the PR? Also please remove package-lock.json.

PR #2029, opened 2017-11-05
Export isSpeakerNotes

lechten:
Export `isSpeakerNotes`, which is useful to avoid multiple audio players, e.g., with the audio-slideshow plugin.


hakimel:
🙌

PR #2030, opened 2017-11-05
Register search key shortcut for help overlay

lechten:
Currently, the search keyboard shortcut does not appear in the help overlay. This PR registers it.


hakimel:
Thanks!

PR #2031, opened 2017-11-06
add multi-column support (incl. readme and demo.html)

nocke:
very simple classes (not a huge grid system) for multi-column support.

selection_178

    <h4>...with the very same classes, you can also have 3 or even 4 columns:</h4>
    <div class='multi-col'>
        <div class='col'>Also works for three columns...</div>
        <div class='col'>...as we can show in...</div>
        <div class='col'>...this example here.</div>
    </div>

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,7 @@ reveal.js comes with a broad range of features including [nested slides](https:/
   - [Parallax Background](#parallax-background)
   - [Slide Transitions](#slide-transitions)
   - [Internal links](#internal-links)
+  - [Multi Column](#multi-column)
   - [Fragments](#fragments)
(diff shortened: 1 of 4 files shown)


bnjmnt4n:
IMO this is a little too specific for Reveal. @hakimel what do you think?

olemis:
Would be nice to have this in markdown as well e.g. something like [remark.js content classes](https://github.com/gnab/remark/wiki/Markdown#content-classes).

olemis:
https://remarkjs.com/#12

3str:
Can I still use the multi-column feature? It seems like this branch was not merge into the current master. What should I do to use the multi-column feature for now? Thanks:-)

Issue #1230, opened 2015-05-13
invalid ID attribute and character stripping


bartoleo:
I'm using a tool to create reveal.js slides from asciidoctor files.
Asciidoctor creates sections with id containing characters outside (a-z,A-Z,0-9,.,-,_,:), e.g.:
à è ì ò ù ...

readUrl functions checks id:

```
// Ensure the named link is a valid HTML ID attribute
            if( /^[a-zA-Z][\w:.-]*$/.test( name ) ) {
```

writeUrl strips charactees:

```
id = id.replace( /[^a-zA-Z0-9\-\_\:\.]/g, '' );
```

These sections are not linkable from url.

There's a reason for these 'checks'?
HTML5 is very permissive about id attribute:
http://www.w3.org/TR/html5/dom.html#the-id-attribute

hakimel:
Happy to make this more flexible than the current HTML 4 validation.

Peque:
[Another consequence of this](1346): if you execute `grunt serve` and modify the HTML file while displaying the slide with a "broken" ID in the browser (imagine a live preview of your presentation while editing it), it will automatically jump to the first slide, instead of displaying the same slide after the edit. Which means after each edit you will have to navigate again to that slide to see the changes.

hakimel:
This was fixed in #2037. The fix is available from the dev branch until next release.

Issue #1346, opened 2015-09-05
Section `id` with special characters


Peque:
Currently, if a section `id` has special characters (i.e.: é, à, ü, ø...) then:
1. The special characters are not displayed in the URL bar. Why?
2. If you execute `grunt serve` and modify the HTML file while displaying that section in the browser, it will automatically jump to the first slide, instead of displaying the same slide after the edit. Is not this a bug?

Content example:

``` html
  <div class="reveal">
    <div class="slides">

<section id="ascii-text">
  <h1>ASCII text</h1>
</section>
<section id="ømq">
  <h1>ØMQ</h1>
</section>

    </div>
  </div>
```

schauder:
At least when an id starts with a digit this will break the navigation. The second slide will be skipped in the following presentation when navigating with the space bar:

        <section>
            <section id="one"><h2>shows</h2></section>
            <section id="2-two"><h2>Gets skipped</h2></section>
        </section>
        <section>
            <section id="three"><h2>Shows to early</h2></section>
        </section>

While this is probably the same issue, it makes it much more grave, since it breaks pretty normal presentations

schauder:
This might be related to #1230

bnjmnt4n:
Closing this as a duplicate of #1230.

PR #2037, opened 2017-11-17
URI encode name hash

nnormand:
Named link derived from slide id attributes with accentuated or special characters can't be found back in the slideshow when reading a URL. For instance the id "lemon-soufflé" produces the hash "lemon-souffl" which, when entered, can't be linked to the correct slide. The fix proposed in this PR consists in URI encoding the id value to form the hash and URI decoding the hash to locate the slide by its id.
This is linked to issue [1346](https://github.com/hakimel/reveal.js/issues/1346) and related issues.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -3563,10 +3563,11 @@
 			var element;
 
 			// Ensure the named link is a valid HTML ID attribute
-			if( /^[a-zA-Z][\w:.-]*$/.test( name ) ) {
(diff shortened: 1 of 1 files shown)


bnjmnt4n:
Seems like a good fix to close #1230 and #1346.

hakimel:
Looks great! This has been merged into the dev branch and will make its way out with the next release.

Issue #2041, opened 2017-11-25
getProgress() incorrect when called from slidechanged event on edge case


denehyg:
The `slidechanged` event is dispatched prior to handling the edge case _where the previous slide maintains the 'present' class when navigating between adjacent vertical stacks_...

https://github.com/hakimel/reveal.js/blob/a0c013606e130baad976195730dcd7e7cd9e2855/js/reveal.js#L2431

When user code listens to the `slidechanged` event and the edge case exists, a call to Reveal.getProgress() returns an incorrect value (as the edge case handling has not yet updated the `present` class on the slides).

In the presentation below, the progress indication for slide 3.1 is **0.25**. When moving to slide 4 it is also **0.25** (should be 0.625).

```
<!doctype html>
<html>
	<head>
		<meta charset="utf-8">
		<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">

		<title>reveal.js</title>

		<link rel="stylesheet" href="https://revealjs.com/css/reveal.css">
		<link rel="stylesheet" href="https://revealjs.com/css/theme/black.css">
	</head>
	<body>
		<div class="reveal">
			<div style="position: fixed; bottom: 10px">Progress: <span id="progress">?</span></div>
			<div class="slides">
				<section>Slide 1</section>
				<section>Slide 2</section>
				<section>
					<section>Slide 3.1</section>
					<section>Slide 3.2</section>
					<section>Slide 3.3</section>
				</section>
				<section>Slide 4</section>
				<section>Slide 5</section>
				<section>Slide 6</section>
				<section>Slide 7</section>
			</div>
		</div>

		<script src="https://revealjs.com/lib/js/head.min.js"></script>
		<script src="https://revealjs.com/js/reveal.js"></script>

		<script>
			Reveal.initialize();

			function updateProgress() {
				document.getElementById('progress').textContent = Reveal.getProgress();
			}
			Reveal.addEventListener( 'ready', updateProgress );
			Reveal.addEventListener( 'slidechanged', updateProgress );
		</script>
	</body>
</html>
```

hakimel:
Fixed in #2042

PR #2042, opened 2017-11-25
Fix for getProgress() bug on slidechanged event

denehyg:
This fixes issue #2041 by dispatching the `slidechanged` event after the vertical stack edge case has been handled.


hakimel:
Merged, thanks!

PR #2045, opened 2017-11-30
Allow tex2jax options to be passed into math plugin.

### LINKED ISSUE #2026: Math plugin: configuration option to disable $ delimiters for inline math
The [math plugin](https://github.com/hakimel/reveal.js/blob/master/plugin/math/math.js#L18)  enables tex2jax and sets `$` as inline math delimiters.  This means that HTML like
```
<p>I made $4 yesterday and $5 today.</p>
```
gets mangled by being interpreted as math.  Could we have a configuration option for this?  Something like:

```js
			tex2jax: {
				inlineMath: options.inlineMath || [['$','$'],['\\(','\\)']] ,
				skipTags: ['script','noscript','style','textarea','pre']
			},
```

or for more control,

```js
			tex2jax: options.tex2jax || {
				inlineMath: [['$','$'],['\\(','\\)']] ,
				skipTags: ['script','noscript','style','textarea','pre']
			},
```

(See also jgm/pandoc#4027.)

jgm:
Closes #2026.

This does not change the default behavior of the math plugin,
but it allows $ delimiters to be disabled (which is usually
what you want, since otherwise normal uses of $ for currency get
treated as math delimiters).

To use:

    Reveal.initialize({
      math: {
        tex2jax: { inlineMath: [['\\(','\\)']],
                   skipTags: ['script','noscript','style',
                              'textarea','pre'] }
            },
      etc.
      });

### DIFF
--- a/plugin/math/math.js
+++ b/plugin/math/math.js
@@ -9,15 +9,15 @@ var RevealMath = window.RevealMath || (function(){
 	var options = Reveal.getConfig().math || {};
 	options.mathjax = options.mathjax || 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js';
 	options.config = options.config || 'TeX-AMS_HTML-full';
+	options.tex2jax = options.tex2jax || {
+				inlineMath: [['$','$'],['\\(','\\)']] ,
+				skipTags: ['script','noscript','style','textarea','pre'] };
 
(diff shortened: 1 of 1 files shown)


reagle:
This will be great when merged, thanks!

hakimel:
Looks great, thanks 🙌

PR #2061, opened 2018-01-02
 Update License Copyright Range

flxwu:
The Copyright should range from the date of licensing (which is commit [7379fb3](https://github.com/hakimel/reveal.js/commit/7379fb3652189e0789ab6b9466616a6f0e461642#diff-9879d6db96fd29134fc802214163b95a)) on _Jun 7, 2011 till the present_.


earboxer:
It should say `2011 - 2020`. That way, when future-someone gets a hold of the source code, they know when the copyright expired.

Issue #1919, opened 2017-06-08
Suggest to add fragments could index


xxssww0258:
My English is not good.
"Fragments" lack "index" parameter, I need to use the "index" of the fragments, similar to the effect of some of the subtitles, continuously Hidden display, fragments
thank you.
我在使用过程中需要做一些类似字幕的效果，但是缺少fragments的index我操作起来不好控制，
例如说我第一次只显示2个fragment ,到第三个的时候隐藏前2个出现5个fragment,每一次都是不同的个数，我需要不断的创建id判断dom是不是等于这个id的dom,
好心人翻译一下，谢谢

bnjmntn:
Duplicate of [#228](https://github.com/hakimel/reveal.js/pull/228#issuecomment-10261098).

PR #2062, opened 2018-01-04
optionally put the fragment in the URL

dougalsutherland:
If you're editing a presentation with fragments, using livereload becomes annoying: after each change, you need to click through the fragments to where you were before. This PR greatly improves that by optionally including the current fragment index in the URL, so that you only have to click around if you're changing the fragment order (similarly to how you have to click around if you add/remove slides).

This is hidden behind a new configuration option `fragmentInURL` (default `false`), since not everyone will want fragments cluttering up their browser history (https://github.com/hakimel/reveal.js/pull/228#issuecomment-10261098).
  
This has been requested a couple of times: #228, #1125, #1919. By putting it behind a configuration option, I think it makes life much nicer for people using livereload while also not changing things at all for those who don't want it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,6 +231,10 @@ Reveal.initialize({
 	// Turns fragments on and off globally
 	fragments: true,
 
+	// Flags whether to include the current fragment in the URL,
+	// so that reloading brings you to the same fragment position
+	fragmentInURL: false,
+
 	// Flags if the presentation is running in an embedded mode,
 	// i.e. contained within a limited portion of the screen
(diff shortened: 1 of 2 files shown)


bnjmnt4n:
@hakimel any opinions on this?

hakimel:
Looks great! Like you mentioned, this has been requested a few times in the past so makes sense to provide an option for it.

I've merged this into https://github.com/hakimel/reveal.js/commits/dev and it'll go out with the next release.

PR #2069, opened 2018-01-11
More Fine-tuned navigation control (horizontal and vertical)

ZadenRB:
Adds a custom navigationEvent, which if cancelled, will prevent a slide change. Also adds functions and config options to toggle either horizontal or vertical navigation being enabled.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -192,6 +192,10 @@
 			// The display mode that will be used to show slides
 			display: 'block',
 
+			//Allow horizontal or vertical navigations
+			horizontalNavigation: true,
+			verticalNavigation: true,
+
 			// Script dependencies to load
 			dependencies: []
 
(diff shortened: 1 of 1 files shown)


bnjmnt4n:
I like the idea of a navigation event that allows slide changes to be cancelled, although maybe it should use `Reveal.addEventListener` like the rest of our API instead of using the DOM event system. However the option to prevent horizontal / vertical navigation doesn't seem very useful to me.

ZadenRB:
@bnjmnt4n, I agree that it should use `Reveal.addEventListener` if possible, I'll look into changing that. The options for horizontal/vertical navigation can be removed/kept in my fork, their main use was for validation (i.e. completing a problem before being allowed to advance) while allowing vertical navigation for hints on the problem.

PR #2080, opened 2018-01-23
Update dependencies.

bnjmnt4n:
This also switches to using QUnit v2, so it might be a pretty significant change. All tests are still passing.


hakimel:
Great stuff. Tested with a fresh install and everything looks 👍

PR #2083, opened 2018-01-25
Add 'Next slide and notes' speaker layout

raboof:
For situations where the current slide is already nicely visible somewhere.


raboof:
Actually this is buggy: when this is the default layout, the 'upcoming slide' window is not correctly initialized anymore, and keeps showing the current slide instead of syncing to the next slide. It works fine when first opening in 'Default' layout and then switching to this one.

Issue #1726, opened 2016-10-13
MathJax not working properly with speaker notes


DL6ER:
Whenever I try to use LaTeX on speaker notes, the output is different from what one would expect: The "raw" input is shown directly behind the correctly rendered math.

Example code:
```html
<aside class="notes">
 <ul>
  <li>Continuous level</li>
  <li>Discrete level. $m,n > 0$, scheme-dependent.</li>
 </ul>
</aside>
```

Result:
![screenshot_speakernotes](https://cloud.githubusercontent.com/assets/16748619/19346204/70a5bde8-9142-11e6-9d4d-2511017f07d3.png)

DL6ER:
Update: The generated HTML source reads:

```html
<span class="MathJax" id="MathJax-Element-9-Frame" tabindex="0" data-mathml="<math xmlns=&quot;http://www.w3.org/1998/Math/MathML&quot;><mi>m</mi><mo>,</mo><mi>n</mi><mo>&amp;gt;</mo><mn>0</mn></math>" role="presentation" style="position: relative;">
<nobr aria-hidden="true"><span class="math" id="MathJax-Span-223" role="math" style="width: 4.071em; display: inline-block;"><span style="display: inline-block; position: relative; width: 3.835em; height: 0px; font-size: 106%;"><span style="position: absolute; clip: rect(1.359em 1003.78em 2.538em -999.997em); top: -2.179em; left: 0.003em;"><span class="mrow" id="MathJax-Span-224"><span class="mi" id="MathJax-Span-225" style="font-family: MathJax_Math; font-style: italic;">m</span><span class="mo" id="MathJax-Span-226" style="font-family: MathJax_Main;">,</span><span class="mi" id="MathJax-Span-227" style="font-family: MathJax_Math; font-style: italic; padding-left: 0.18em;">n</span><span class="mo" id="MathJax-Span-228" style="font-family: MathJax_Main; padding-left: 0.298em;">&gt;</span><span class="mn" id="MathJax-Span-229" style="font-family: MathJax_Main; padding-left: 0.298em;">0</span></span><span style="display: inline-block; width: 0px; height: 2.185em;"></span></span></span><span style="display: inline-block; overflow: hidden; vertical-align: -0.247em; border-left-width: 0px; border-left-style: solid; width: 0px; height: 1.066em;"></span></span></nobr>

[...]

PR #2090, opened 2018-02-08
Allow users to customise MathJax options.

bnjmnt4n:
Ref. #1856, #2006, #2045.

This is a more open approach to allow customisation of all MathJax options instead of select options only.


rschmehl:
Seems to be a useful improvement for the use of MathJax in RevealJS and I propose to also take along a simple fix for the use of MathJax in speaker notes. See https://github.com/hakimel/reveal.js/issues/1726#issuecomment-363824328

PR #2096, opened 2018-02-14
Background repeat & position to parallax background

Jmuccigr:
Additional background features, especially useful for those who don't use parallax.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -173,6 +173,12 @@
 			// Parallax background size
 			parallaxBackgroundSize: '', // CSS syntax, e.g. "3000px 2000px"
 
+			// Parallax background repeat
+			parallaxBackgroundRepeat: '', // repeat/repeat-x/repeat-y/no-repeat/initial/inherit
+
+			// Parallax background position
+			parallaxBackgroundPosition: '', // CSS syntax, e.g. "top left"
+
 			// Amount of pixels to move the parallax background per slide step
 			parallaxBackgroundHorizontal: null,
 			parallaxBackgroundVertical: null,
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks for the PR. Can you provide an example parallax configuration that shows what these values would be used for?

I never really use parallax backgrounds myself but I figured the goal of it is for the image to cover the whole screen and scroll slightly with each slide change. That said I'm curious about how it would work with a repeating image.

G1305:
sorry @hakimel for worrying,
I would like DYNAMIC slides plugin for add included slides (click to open only)
I hope this is very important for everyone... 
THANK YOU

Jmuccigr:
@hakimel, I believe that the default is to repeat the parallaxBackgroundImage when it's smaller than the window, which may not be desirable. For example, you might want your image centered, or have it as a logo in a corner. So you need to be able to set the repeat and the location for that.

I tend to use a background image, but leave the parallaxBackgroundHorizontal and vertical at null, so that it doesn't move; it just gives me a nice background. You can already set the movement to be something other than null and reveal.js will respect that. I generally don't want a repeating image though.

(Is that what you meant?)

hakimel:
I see. So you're essentially using the parallax background settings to set a fixed background image? The parallax fields were introduced specifically for the parallax/moving effect so they are needlessly complex for adding a regular background image.

I like the idea of being able to set a presentation-wide background via configuration options. I'm going to merge your PR but in the future I'm going to rename the configuration fields so make it clear they can be used for non-parallax backgrounds too (`parallaxBackgroundImage` -> `backgroundImage`, `parallaxBackgroundRepeat` -> `backgroundRepeat` etc).

Jmuccigr:
Either way is fine with me.

Note that the horizontal and vertical values are set to null by default in reveal.js, so that you get a non-moving image unless you specify otherwise. It might be simpler just to leave it as is, or change the variable's name to drop the "parallax" bit, unless the extra code is burdensome. Then users can put in a background image, and move it simply by setting the H and V values without worrying about setting a separate parallax background image.

PR #2104, opened 2018-02-20
Made the speaker's notes usable without the server

jurca:
Currently it is necessary to serve the slides over HTTP for the speaker's view to work.

This patch makes the speaker's view work even when the slides are opened from the filesystem.


valencik:
Thank you for this, it works wonders.

hakimel:
Thanks @jurca! These changes have been merged in now.

PR #2114, opened 2018-03-07
Prefer W3C pointer events and updates to match the standard definition

ChumpChief:
The [W3C Pointer Events standard](http://www.w3.org/TR/pointerevents/) omits navigator.pointerEnabled, so this method of detecting pointer support won't work on standard-compliant browsers ([Edge, Chrome](https://caniuse.com/#feat=pointer)).  This PR updates to check for 'onpointerdown' in window instead, and removes conditionality on event removal (since it is safe to removeEventListener even when the event is not registered).  Doing so enables swipe navigation to work on Edge.

This PR also reorders the conditional to prioritize use of pointer events (when available) over touch events.  Since pointer events don't necessitate main-thread handling due to event registration alone, this opens opportunities for slide authors to restrict swipe behavior from portions of their content and retain off-thread panning performance for overflow content if they desire, via scoped application of the CSS touch-action property.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -1172,13 +1172,8 @@
 		window.addEventListener( 'resize', onWindowResize, false );
 
 		if( config.touch ) {
-			dom.wrapper.addEventListener( 'touchstart', onTouchStart, false );
-			dom.wrapper.addEventListener( 'touchmove', onTouchMove, false );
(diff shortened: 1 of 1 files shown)


hakimel:
Great update–thanks!

PR #2115, opened 2018-03-07
Replace toArray() with ES6 Array.from 

DaniPani:
I tried to replace `toArray` function with `Array.from`. Everything seems to work fine. Except for grunt test because PhantomJs doesn't support ES6. It might a good option


DaniPani:
Ops :(

PR #2117, opened 2018-03-08
Replace toArray with ES6 Array.from

DaniPani:
I have replaced `toArray` function ES6 `Array.from`. Everything works fine, expect for grunt test because PhantomJS doesn't support ES6


bnjmnt4n:
Hi @DaniPani! We currently support all ES5 browsers, so adding ES6 would cause older versions of some browsers to fail, for example PhantomJS. Closing this for now.

PR #2120, opened 2018-03-10
Added overlay to speaker notes upcoming slides to prevent mouse focus

denehyg:
This adds an overlay to the upcoming slides notes to prevent mouse clicks from taking input focus away from the main speaker notes window. Currently these mouse clicks change input focus directly to the upcoming presentation and subsequent keyboard events are isolated to these slides only (rather than the presentation itself).

### DIFF
--- a/plugin/notes/notes.html
+++ b/plugin/notes/notes.html
@@ -34,6 +34,14 @@
 				z-index: 2;
 			}
 
+			/* prevent upcoming slides iframe from taking mouse focus */
+			#upcoming-slide .upcoming-overlay {
+				background-color: transparent;
+				width: calc(100% - 12px);
+				height: calc(100% - 12px);
+				position: absolute;
+			}
+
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks for the PR but this is intentional. Clicking and navigating inside of the upcoming slide allows you to look ahead without showing your audience. In the slides.com speaker view we also added a button that allows you to jump to the upcoming slide and this is something I want to add to bring over to reveal.js as well.

denehyg:
Understood. But I think this presents a UX issue. When the presenter accidentally/unknowingly clicks in the upcoming slide view then later tries to advance using the keyboard they are expecting the presentation to advance. This can be quite distracting when in the middle of a presentation.

Perhaps there needs to be a clear indication that the upcoming slides view has keyboard focus (or that the rest of the speakers notes do not).

PR #2121, opened 2018-03-10
add support for overriding the default layout

SteveHartzog:
New `overrideLayout` option (if true) will prevent h/w calcs.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -346,7 +346,7 @@
 		if( typeof query['dependencies'] !== 'undefined' ) delete query['dependencies'];
 
 		// Copy options over to our config object
-		if( options.overrideLayout ) config.display = '';
+		if( typeof(options) === 'object' && options.hasOwnProperty('overrideLayout') && options.overrideLayout ) config.display = '';
 		extend( config, options );
 		extend( config, query );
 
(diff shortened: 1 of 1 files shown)


js/reveal.js
```diff
@@ -342,6 +346,7 @@
 		if( typeof query['dependencies'] !== 'undefined' ) delete query['dependencies'];
 
 		// Copy options over to our config object
+		if( typeof(options) === 'object' && options.hasOwnProperty('overrideLayout') && options.overrideLayout ) config.display = '';
```
hakimel: Can we remove this? Feels a little unexpected that disabling the layout is mutually exclusive with setting the display style style.

hakimel:
Looks like a useful addition, thanks!

I wonder if `disableLayout` is a more appropriate name. The config option disables the default layout logic, it doesn't really override it in itself.

js/reveal.js
SteveHartzog: Yeah, it didn't specifically impact the override... I just didn't want to see it added. :)

SteveHartzog:
Sounds good to me, I'll update my pull to `disableLayout`. and remove the `display` bits. Thanks for reviewing so quickly.

hakimel:
Thanks for updating!

PR #2122, opened 2018-03-11
Added Gulp

DaniPani:
I have created a `gulpfile.js` with the same tasks of grunt, replaced `npm script` and finally added `.jshintrc`


Delapouite:
Thanks for this PR. I think `eslint` support would be more appropriate than `jshint`.

DaniPani:
And voila!

pndewit:
I know this is very old, but may I ask why Grunt is still in the PR? I would expect Grunt to be replaced by Gulp instead of living side-by-side in this PR.

Also, I see NPM was added as a dependency. I don't think it is a good practice to have NPM as a dependency because of multiple reasons. For example, what if the user wants to use Yarn?

hakimel:
This has been merged, thanks!

I made a couple of changes to the tasks and still need to get the tests working. I'll also be removing grunt, so as of reveal.js 4.0 gulp will be the way to go 🙌

Issue #1686, opened 2016-09-04
[Workaround] White background when printing pdf


ebonet:
I was trying to print a pdf from my presentation using the moon theme, but it always printed a white background instead of the background color itself, even though chrome showed the background with correct colors. 

I figured the problem was with the css: the bg color was applied to the body, but never to the .slide-background class (it was transparent). My work around was to add the following piece of code to my custom css. A better solution is to add it on the scss itself.

``` css
@media print {
    .slide-background {
        background-color: #002b36 !important;
        -webkit-print-color-adjust: exact; 
    }
}
```

Hope it helps someone

hakimel:
Fixed in #2131

grasshopyx:
It doesn't work for me. But the #1686 workaround works.

tbrodbeck:
Did not work for me either. Please reopen.

PR #2131, opened 2018-03-19
Print background in pdf

quentinproust:
Pull request to integrate issue https://github.com/hakimel/reveal.js/issues/1686. 
Adding the workaround to the theme template to make it work for every theme.

### DIFF
--- a/css/theme/template/theme.scss
+++ b/css/theme/template/theme.scss
@@ -316,9 +316,8 @@ body {
 /*********************************************
  * PRINT BACKGROUND
  *********************************************/
-@media print {
-    .reveal .slide-background {
-        background-color: $backgroundColor !important;
-        -webkit-print-color-adjust: exact; 
+ @media print {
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks for the PR. Would be great to fix this – I've never noticed this since I've always checked the "Background graphics" when printing in Chrome. 

Can you check if the following fix works for you as well? It's cleaner if we can avoid !important overrides on a per-slide background basis. `-webkit-print-color-adjust` isn't needed as we already set that for the whole document in the PDF print sheet.

```
@media print {
    .backgrounds {
        background-color: $backgroundColor;
    }
}
```

quentinproust:
@hakimel Thanks for your suggestion. I just applied the workaround from issue #1686 without looking futher.
Your suggestion worked for me. I updated the pull request.

hakimel:
Great, thanks!

Issue #2067, opened 2018-01-10
Slide numbers don't match URL 


Leftium:
Apparently, the URL is 0-indexed, but the slideNumber starts at 1. It would be less confusing if they used the same numbering system.

**Repro steps:**

1. Add option `slideNumber: true` to https://github.com/hakimel/reveal.js/blob/master/demo.html
2. Compare URL/slide number in browser

![image](https://user-images.githubusercontent.com/381217/34772863-e7f3017e-f64c-11e7-9fb0-38a7c8b306e7.png)

bnjmnt4n:
It is true that using the same numbering system is less confusing, however the 0-indexed URL system has been in use since the very first version of Reveal and hence it is unlikely that we will switch to a different system.

Leftium:
What about changing the displayed slide numbers to also be 0-indexed?

PR #2133, opened 2018-03-22
Adding support for hash linked slideNumber

sean-parent:
There are two changes here -

Make the hash slide numbering match the shown slide so that `slide.html#/3` is slide 3 and not slide 2.

Added an href link to the slide number if shown to make it simple to share a link to a particular slide even when history is not enabled. This also makes it easy to lock to a particular slide when using auto refresh tools while developing.


mschilli87:
This was suggested before but decided against: https://github.com/hakimel/reveal.js/issues/2067

sean-parent:
Recommendations to proceed?
I could make 1 based hash indexing optional
I could drop 1 based hash indexing and just keep the href link in the slide number
I could make the href link in the slide number optional
Other suggestions?

hakimel:
I think it'd be best to make the 1-based indexing optional to avoid breaking existing content. If you can revise that I'd be happy to merge this in. Thanks!

sean-parent:
One based indexing is now optional (defaults to zero based for compatibility)

hakimel:
Thanks for updating. One more thing, can you change the merge target of the PR to the 'dev' branch? That's the working branch, we merge dev into master when a new version goes out.

marcosalpereira:
There is still something preventing the incorporation of this change?

mschilli87:
@hakimel: If it's just the wrong target branch, you should be able to change that yourself:
https://blog.github.com/2016-08-15-change-the-base-branch-of-a-pull-request/

sean-parent:
Merged to dev branch and resolved conflicts.

hakimel:
Thanks for resolving the conflicts 🙌

PR #2141, opened 2018-04-01
Clarify the meaning of the notes?: regex

marcvangend:
When I read that the default notes separator in markdown files is `notes?:`, I took it too literally and put `notes?:` in my markdown, which of course did not work. Unfortunately it took me a debugger to realize why I was wrong.

Going back to the documentation, I noticed that it does say that the notes separator is interpreted as a regex. Although the documentation is not wrong, it didn't help me understand what the default separator actually means. Hopefully this pull request will prevent others from making the same mistake.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,7 +125,7 @@ This is based on [data-markdown](https://gist.github.com/1343518) from [Paul Iri
 
 #### External Markdown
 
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks!

Issue #2162, opened 2018-05-18
Reveal.js highlighting is broken on slow internet connection and/or server being slow


Manvel:
## Background
I did notice this issue when I was preparing [current](https://cmints.io/en/presentation#/) presentation and experiencing slow response from my server. Anyway I have prepared actual reproduction steps with the default Reveal.js install below:

## Reproduction steps
- Clone [revealJS](https://github.com/hakimel/reveal.js) repository and run the server
- Replace Slide1 with content below:
```html
<section>
  <h3>Slide 1</h3>
  <pre><code>
  const date = new Date(2018, 10, 12);
  console.log(new Intl.DateTimeFormat('en-GB').format(date));
  console.log(new Intl.DateTimeFormat('en-US').format(date));
  console.log(new Intl.DateTimeFormat('lt-LT').format(date));
  </code></pre>
</section>
```
- Navigate to http://localhost:8000/ in Chrome Browser (Were using Chrome v66 on Mac)
- Open Chrome Dev Tools and from the Network tab select `Slow 3G preset ` 
- Clear the cache and reload the page
- Observe

### Observed Behavior

Highlight.js doesn't highligh `<pre><code>` blocks and current error is thrown in the console:

```
Uncaught ReferenceError: hljs is not defined
    at callback ((index):52)
    at reveal.js:423
    at f (head.min.js:6)
    at Array.<anonymous> (head.min.js:6)
    at u (head.min.js:6)
    at head.min.js:6
    at e (head.min.js:6)
    at head.min.js:6
```

### Expected behavior
No error is thrown and code is highlighted after the highlight.js is downloaded.

## Notes:
- I did notice that reveal.js using some version of highligh.js which is almost 500kb(I assume because it contains all languages). The suggested cdnjs version(with common languages) in https://highlightjs.org/download/ is less than 5kb.
- Even though I think this issue is connected with how Reveal.js is dealing with the dependency scripts updating the library might be a good start for fixing this issue.

Manvel:
- I just created a [pull request](https://github.com/hakimel/reveal.js/pull/2163) to replace big highligh.js file, this fix will also resolve the reproduction steps in current issue, but I still think that it make sense to check why Reveal.js couldn't handle this in the [dependencies loading implementation](https://github.com/hakimel/reveal.js/blob/master/index.html#L40) (I still think this issue might strike back after again)
- Also do you know, why the old version of the highlightJS was so big ?
  - Did it include more languages ?

[...]

PR #2163, opened 2018-05-18
#2162 - Replaced highlightJS 9.11 with 9.12(prebuilt with 23 commonly used languages)

Manvel:
I did notice that reveal.js using some version of highligh.js which is almost 500kb(476kb), while the suggested cdnjs version in https://highlightjs.org/download/ is less than 5kb.

See #2162 for more details.


rugk:
@hakimel Any news here?

hakimel:
Thanks for the suggestion but I prefer leaving all of the languages in by default for ease of use. Smaller file size isn't a high priority since reveal.js is primarily used to present from your own local machine.

I agree it's a inconveniently large file when your intention is to host the presentation. In that case you could build an optimized highlight.js version that contains only the language(s) you need.

PR #2172, opened 2018-06-05
learning git 

yamacode:
Reveal.js was used an as example to fork and then create a pull request


rugk:
Does not seem as if the changes should really be merged into the master though. So if this is the case, you can close this PR.

PR #2199, opened 2018-07-23
Fix: Suporting unicode id's

### LINKED ISSUE #2198: unicode in section ids do not reload properly
With `history=true` reloads keep the slide, which is quite nice when you are editing them. But when the id has some unicode character within, reveal rewrites the url removing such characters from the url hash and reloads to the first slide. This is quite nagging when you are writing a slide and want to see the effects of your changes you just made.

When the id is set to some value, is quite pointless [to set it to a different one](https://github.com/hakimel/reveal.js/blob/496db06441f344f3afeb00c2d62f734f44a9e3e4/js/reveal.js#L3747) as it won't go anywhere. Html5 allows "any non-space unicode character" as id. If authors use illegal ids on html4 it's on them.

The need comes from the fact that Pandoc generates section id's from titles, and some of them have such unicode characters. I could ask Pandoc the feature of removing such chars from the id, but as i understand the proper fix is on revealjs.

This is related to #1655.

vokimon:
Reveal should'nt care about the actual content of the slide id (filtering chars which are not ascii letters, numbers or some simbols). HTML4 cares, but html5 does not.

Even, in html4, there is no point in manufacturing a new id to a target that does not exist. If searching for unicode ids does not work, the effect would be the current behavior (going back to the first page).

What it is needed is to urldecode the given hash.

It fixes  #2198. About failing to reload slides with Pandoc generated ids, which are based on titles that might contain unicode.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -3691,12 +3691,8 @@
 		// assume that this is a named link
 		if( isNaN( parseInt( bits[0], 10 ) ) && name.length ) {
 			var element;
-
-			// Ensure the named link is a valid HTML ID attribute
-			if( /^[a-zA-Z][\w:.-]*$/.test( name ) ) {
-				// Find the slide with the specified ID
-				element = document.getElementById( name );
-			}
+			name = decodeURIComponent(name);
+			element = document.getElementById( name );
 
 			if( element ) {
 				// Find the position of the named slide and navigate to it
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks for the PR but this had already been addressed by someone else in fa8a7334ce2ea99a736dd5e72734b08fc01c6fdca. The change was in the dev branch, but I've released it in reveal.js 3.7.0 now. Please give it a shot and let me know if you encounter any issues.

vokimon:
Super. Works like a charm. Even better than my patch. Thanks for the release!

PR #2208, opened 2018-08-04
Use 'background: initial' on img.plain

felixfischer:
This change removes the inherited translucent white background from `.reveal section img.plain`. I think there should be no special background on `.plain` images, as this causes display problem with transparent SVG logos on non-white backgrounds.


felixfischer:
closing this as it is deprecated for me

PR #2209, opened 2018-08-06
Merge pull request #1 from hakimel/master

shgninc:
update from origin


shgninc:
update to new releases from origin

PR #2215, opened 2018-08-15
Only show resume on pause if controls enabled

anderslemke:
Hi @hakimel ,

This is such an awesome project! Thank you for making it 👏 

I really love the idea of the pause-mode.
I'm not a big fan of the "Resume presentation"-button though... 🤭

I tried figuring out why it was introduced, but I couldn't.

So in this PR I'm proposing two things:

1. Use `config.controls` to control whether or not the button is shown.
2. Let the entire `dom.pauseOverlay` listen for clicks and close when clicked.

This is just one way you could approach this. You could also introduce a new config for the button, but I thought it made good sense to use the `controls`.

Let me know what you think.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -593,8 +593,7 @@
 		dom.speakerNotes.setAttribute( 'tabindex', '0' );
 
 		// Overlay graphic which is displayed during the paused mode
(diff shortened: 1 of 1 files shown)


hakimel:
Both changes make sense – thanks!

The resume button was added because some people were accidentally triggering the pause mode and getting stuck without knowing how to disable it.

anderslemke:
Ah, yes of course. That makes sense. "Where the 🤬 did my presentation just go?" 🤪

PR #2231, opened 2018-09-18
bring install instructions to top, as is tradition

esthor:
The Install was waaaay at the bottom, which didn't make sense to scroll through all the detailed documentation, just to find how to get started with an install.


hakimel:
Makes sense, thanks!

PR #2234, opened 2018-09-21
Cut and pasted my slides in.

jayseeg:


jayseeg:
I'm in the wrong place,... my bad

Issue #1952, opened 2017-07-30
Simplifying the CSS of Reveal.js?


rauschma:
I absolutely love Reveal.js. The only thing giving me trouble was customizing the CSS. These are a few examples (I’ll conclude with suggestions for improvement):

At the moment, some of Reveal’s CSS is very dominant. For example, keywords in code (highlighted via highlight.js) are only bold if you force the issue via !important:

```css
.hljs-keyword,
.hljs-selector-tag,
.hljs-subst {
  color: #333 !important;
  font-weight: bold !important;
}
```

You see a similar issue in pdf.css which has many occurrences of !important: https://github.com/hakimel/reveal.js/blob/master/css/print/pdf.css
That’s a problem, because it overrides some other CSS and undoes customizations that work in non-PDF mode.

Summary:

* At the moment, Reveal’s CSS is relatively difficult to customize. (Caveat: I’m saying this as someone whose knowledge of CSS is limited.)
* If one relied on modern CSS features (esp. CSS), could it be restructured and simplified so that !important is needed less (or not at all)?
* Could one of the themes *not* have text-align: center (which does not work well on slides with more than one bullet list)? I’m overriding it as follows in [my custom theme](https://gist.github.com/rauschma/abdeccdfe1bd3946b2361c803db2b65a) (feedback welcome), but I’m not sure I’m doing it right.

```css
.reveal .slides {
  text-align: left;
}
.reveal .slides h1, .reveal .slides h2 {
  text-align: center;
}
```

SebJansen:
Any update on this, @hakimel?

hakimel:
Yeah – plenty of those overrides are no longer needed and could be removed. Some of the style resets, like `text-shadow: 0 0 0 #000 !important`, could probably be removed altogether.

I don't mind the use of !important on the .reveal and .slide containers since those are all related to laying out content so that it can be printed, but overriding specific element styles (like the pre code font-family) is limiting.

rugk:
BTW for such style resets you mention (that are used to make the visibility the same in multiple browsers) there exist various small CSS stylesheets that are explicitly created for this job.

hakimel:
I wasn't referring to normalizing the browser's default styles, we already use reset.css for that. I meant reset as in undoing the styles that reveal.js has set.

albedoa:
[...]

PR #2248, opened 2018-10-03
Decrease specificity of style resets

heavywatal:
Leading `.reveal` has been removed from style resets at the top of `reveal.scss`. I hope this PR will be a first step to address the issue #1952.

### DIFF
--- a/css/reveal.css
+++ b/css/reveal.css
@@ -8,19 +8,19 @@
 /*********************************************
  * RESET STYLES
  *********************************************/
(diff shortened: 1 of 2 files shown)


hakimel:
Thanks for the PR. I've addressed this in a slightly different way in 6abc6e00581e66690416978de118145e854c3c1e

By removing the .reveal prefix we're affecting styling of the page as a whole which might cause styling conflicts with the surround page. Because of that I think it's better to move the reset styles to a separate file so that you can easily leave them out if you want to.

PR #2251, opened 2018-10-05
Give focus to previously opened speaker notes (fixes #2229)

### LINKED ISSUE #2229: Speaker notes timer stops working when opened twice
To reproduce:
1. Press S to open the speaker view (timer works)
2. Press S to open the speaker view again while it's still open (timer stops)

Discovered in #1922

oyron:
Handle previously opened notes window, by giving focus rather than reopen.
Fixes #2229

### DIFF
--- a/plugin/notes/notes.js
+++ b/plugin/notes/notes.js
@@ -11,7 +11,14 @@
  */
 var RevealNotes = (function() {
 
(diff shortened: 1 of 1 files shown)


hakimel:
This is great. Thanks for the fix!

PR #2259, opened 2018-10-16
Update README.md

Kasiban:
changedminimum required version of node.js from version 4.0.0 to 9.0.0


hakimel:
Thanks for the fix, this was updated in #2499

PR #2266, opened 2018-10-30
Make code font customizable for themes.

f00ale:
This patch makes it possible to customize the font used for code.
If nothing is specified the old default monospace font is used.


hakimel:
Makes sense, thanks!

PR #2274, opened 2018-11-08
Print pagesize fix

keflavich:
I had an issue where each page was growing to be huge, so the n'th page was n-times larger than the 1st page.  Removing the multiplication by page size fixed the issue for me.  I'm not entirely sure this is a generally correct solution, though, as for some of my presentations, it makes the slides appear in browser all stacked on top of one another.

I believe the issue is related to setting a background-image for the slide.  With a background image included, the multiplication doesn't work, while without a background image, maybe it does?  I'm open to better suggestions on how to fix this issue in general, but I want to at least raise this as an issue with a possible short-term/workaround solution.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -738,7 +738,7 @@
 				// so that no page ever flows onto another
 				var page = document.createElement( 'div' );
 				page.className = 'pdf-page';
-				page.style.height = ( ( pageHeight + config.pdfPageHeightOffset ) * numberOfPages ) + 'px';
+				page.style.height = ( ( pageHeight + config.pdfPageHeightOffset )  ) + 'px';
 				slide.parentNode.insertBefore( page, slide );
 				page.appendChild( slide );
 


keflavich:
This is probably a wrong fix, I don't think I understand well enough how browsers handle page sizes, page breaks, etc.

The problem I'm having is that I have been including images using things like:
```
<section>
    <span style="background-image:url(image.png); margin-top:-2%; height:100%">
</section>
```
which do not render correctly in print-pdf mode.  By forcing `min-height: 768px`, for example, I was able to force these to render _reasonably_, but not _correctly_.  As is, they render as 0-px height and therefore are not displayed at all.

cujarrett:
@keflavich - Do you have a presentation that demonstrates the problem I can pull down to look at?

keflavich:
The talks on this page (warning: they're pretty big):
https://keflavich.github.io/talks/
Source:
https://github.com/keflavich/talks
note that reveal.js in this directory is patched as in this PR

hakimel:
Hey! When the contents of a slide is taller than the printed page height, we allow the content to span multiple pages. The change in this PR would break that behavior.

That said I'd love to figure out and fix the issue that you've run into. I tried looking at your existing presentations but there's a lot of content in there and I'm not sure whether it's working correctly or not. Any chance you could provide a reduce test case with just a few slides in a clean copy of reveal.js?

keflavich:
Next time I'm working on a reveal talk, I'll post something here.  Maybe I can take some screenshots with/without this modification to `reveal.js` to show the problem more clearly.

PR #2279, opened 2018-11-18
Issue2243

### LINKED ISSUE #2243: Slide Notes timer doesn't work
When I open the Slide Notes browser page (pressing s) the timer doesn't start. I'm using Chrome and I'm getting this error in the console

```javascript
Uncaught ReferenceError: Reveal is not defined
    at getTimings (notes.html:517)
    at setupTimer (notes.html:574)
    at handleConnectMessage (notes.html:413)
    at notes.html:380
```
My speech and plugin are not in the same domain

You can [access the address](https://www.caijw.com/reveal/1613107875995952125.html) to reproduce the problem

sandy8111112004:
fix #2243 

1. includes reveal cdn in notes.html
2. fix function typo from getSlides to getSlide

### DIFF
--- a/plugin/notes/notes.html
+++ b/plugin/notes/notes.html
@@ -2,9 +2,8 @@
 <html lang="en">
 	<head>
 		<meta charset="utf-8">
-
 		<title>reveal.js - Slide Notes</title>
-
+		<script src='https://cdn.jsdelivr.net/reveal.js/2.6.2'></script>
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks but `getSlides()` is the right method in this context. The issue this is trying to fix has already been fixed in another way (see my comment in #2243).

PR #2280, opened 2018-11-20
Fix: Ctrl + click for zoom instead of Alt + click in Demo

tv3141:
Here is the current demo slide explaining zoom: https://revealjs.com/?transition=slide#/4
For me zoom only works with `Ctrl` + click, instead of `Alt`+ click as explained in the demo.

### DIFF
--- a/demo.html
+++ b/demo.html
@@ -93,7 +93,7 @@
 						Press <strong>ESC</strong> to enter the slide overview.
 					</p>
 					<p>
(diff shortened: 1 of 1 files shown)


hakimel:
Alt + Click is the right shortcut, but we use Ctrl + Click on Linux systems. This was changed in a PR a few years back because someone discovered the Alt + Click didn't work in Linux.

Sorry about the confusion. I've updated that slide in the demo presentation 82076c5a0b1aebe7acd0db1886be9dd6445fa68f

PR #2286, opened 2018-12-03
Add new 'hash: true' option which uses replaceState for url

### LINKED ISSUE #2211: Feature: Additional hash / history settings
## current

Currently there are two options which behave as follows:

### `history: false` (this is the default behaviour)

- [ ] slide change adjusts hash
- [ ] slide change pushes to history stack
- [x] loading page with hash takes you to that slide

### `history: true`

- [x] slide change adjusts hash
- [x] slide change pushes to history stack
- [x] loading page with hash takes you to that slide

## desired

I'd like to be able to have the url update without pushing to the history stack, essentially `replaceState()` instead of `pushState()` (it's currently just using `window.location.hash` assignment which is ~roughly equivalent to `pushState`

Here's my desired settings (feel free to bikeshed on the new property name)

### `history: false`, `hash: false` (this is the default behaviour)

(same as `history: false` before)

- [ ] slide change adjusts hash
- [ ] slide change pushes to history stack
- [x] loading page with hash takes you to that slide

### `history: true` (implies `hash: true`)

(same as before)

- [x] slide change adjusts hash
- [x] slide change pushes to history stack
- [x] loading page with hash takes you to that slide

### `history: true`, `hash: false`

This is an illegal combination, `history: true` *implies* `hash: true`.  (technically I think `pushState` supports adding history states without modifying the url so this could be supported, but I don't think it's worth thinking about?).

### `history: false`, `hash: true`

(this is the thing that I want)

- [x] slide change adjusts hash
- [ ] slide change pushes to history stack
- [x] loading page with hash takes you to that slide

## addendum / thanks

- thanks for the project and to everyone who has contributed, really awesome stuff.  I'm currently using it heavily in my [`markdown-to-presentation`](https://github.com/anthonywritescode/markdown-to-presentation) project which converts simple markdown to a reveal.js presentation
- I'd be happy to hack on this if this is seen as a desirable feature and [usage of `replaceState` / `pushState`](https://caniuse.com/#feat=history) fits whatever browsers the project currently supports
- here's some [prior art of mine](https://github.com/tonybaloney/pep-explorer/pull/2/files) which uses `replaceState` / `pushState` with hashes -- [and in action](https://tonybaloney.github.io/pep-explorer/)

asottile:
Resolves #2211 

(this diff may be easier to view in [whitespace ignored mode](https://github.com/hakimel/reveal.js/pull/2286/files?w=1))

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,8 +259,11 @@ Reveal.initialize({
 	// Display the page number of the current slide
 	slideNumber: false,
 
-	// Push each slide change to the browser history
+	// Push each slide change to the browser history.  Implies `hash: true`
 	history: false,
+
+	// Change the hash when changing slides -- impacts browser history with `history: true`
+	hash: false,
 
 	// Enable keyboard shortcuts for navigation
 	keyboard: true,
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -76,8 +76,11 @@
 			// Determine which displays to show the slide number on
 			showSlideNumber: 'all',
 
-			// Push each slide change to the browser history
+			// Push each slide change to the browser history.  Implies `hash: true`
 			history: false,
+
+			// Change the hash when changing slides -- impacts browser history with `history: true`
+			hash: false,
 
 			// Enable keyboard shortcuts for navigation
 			keyboard: true,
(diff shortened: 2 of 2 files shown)


hakimel:
This is great, thanks!

PR #2295, opened 2019-01-06
Quick workaround for .stretch in overview mode and issues with browsers.

gw0:
Alternative workaround for #1187.


hakimel:
I think the fix in #1188 made sense so I've merged that instead. Thanks for providing another option though.

PR #2300, opened 2019-01-15
Resize Event

mw75:
### Resize Event

When reveal.js changes the scale of the slides it fires an resize event. You can subscribe to the event to resize your elements accordingly.

```javascript
Reveal.addEventListener( 'resize', function( event ) { /* console.log(event.scale,event.oldscale,event.size); */ } );
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,7 @@ reveal.js comes with a broad range of features including [nested slides](https:/
   - [Fullscreen mode](#fullscreen-mode)
   - [Embedded media](#embedded-media)
   - [Stretching elements](#stretching-elements)
+  - [Resize Event](#resize-event)
   - [postMessage API](#postmessage-api)
 - [PDF Export](#pdf-export)
 - [Theming](#theming)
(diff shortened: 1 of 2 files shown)


dougalsutherland:
I was trying to do this with a handler on the `window` resize event and using `Reveal.getScale()`, but had trouble because (I think) it sometimes fired before the reveal resize handling. So this would have been helpful for me. (I figured out another way to do what I wanted, though.)

Note though that the name might conflict with [this proposal](https://drafts.csswg.org/resize-observer-1/) for element `resize` events.

README.md
```diff
@@ -958,6 +959,14 @@ Limitations:
 - Only direct descendants of a slide section can be stretched
 - Only one descendant per slide section can be stretched
 
+### Resize Event
+
+When reveal.js changes the scale of the slides it fires an resize event. You can subscribe to the event to resize your elements accordingly.
```
dougalsutherland: `an` should be `a` here

hakimel: Good catch, just merged this and will fix the typo.

PR #2305, opened 2019-01-20
Upgrade marked to 0.6.0

sestegra:
Many fixes have beed found since the embedded version (see https://github.com/markedjs/marked/releases).


tasn:
Yes, please! It also supports github flavoured markdown natively, and allows mixing html and markdown.

It's just a drop-in replacement, I'm using the latest version here (with gfm: true in the configuration).

sestegra:
@tasn Markdown in html doesn't work with this new version.
``` html
<div>
# Test
- A
- B
- C
</div>
```

After launching `marked test.md`, no change on the result is shown.

tasn:
@sestegra it does.

1. You need to enable github flavoured markdown from the configuration.
2. You need an empty line between the markdown and the tags (as per the gfm spec).

Works for me.

hakimel:
Merged into the dev branch, thanks!

PR #2307, opened 2019-01-21
Enable simpleNavigation

sestegra:
Most of remote control for presentation only has next/previous button.

To allow the logical progress to be followed with this kind of hardware, arrows should trigger only next and previous slides.

`simpleNavigation` configuration parameter would be used for that purpose.

I suggest to mix `gridNavigation` and  `simpleNavigation` configuration parameters into `navigationMode` to select from one fo following navigation mode:
- normal (default)
- simple (this pull request)
- grid

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -108,6 +108,9 @@
 			// Change the presentation direction to be RTL
 			rtl: false,
 
+			// When this is enabled, stepping only with previous and next slide.
+			simpleNavigation: false,
+
 			// When this is enabled, stepping left/right from a vertical stack
 			// to an adjacent vertical stack will land you at the same vertical
 			// index.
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks for the PR. I like the idea of a 'navigationMode' config value that incorporates all three of these navigation settings. Gonna take a look at how to best implement it that way.

hakimel:
I've made a number of changes to this:
1. All options are now part of a new 'navigationMode' config value as suggested.
2. I renamed 'simple' to 'linear' to try to convey what changes.
3. When `linear` is used, the vertical arrows are automatically hidden.
4. The visible control arrows on screen now follow the same behavior, so clicking the right arrow when 'linear' is enabled will also step through vertical slides.

Here's the documentation:
https://github.com/hakimel/reveal.js/blob/dev/README.md#navigation-mode

sestegra:
@hakimel Keyboard shortcuts table should be updated to display the relevant actions regarding the navigation mode. I created #2313 issue to follow it.

ShKlinkenberg:
I'm not getting navigationMode: 'linear', to work.

PR #2311, opened 2019-01-29
Modularized slides with example.

twinkybot:
Gulpfile is most probably not the best (didn't know how to do this with grunt)
Executing 'npm buildSlides' will create within build folder a new html file named after the variable in slideDeck.js


hakimel:
Thanks for the PR. I like the idea of this, but I think it's something that should live in a separate repo outside of reveal.js. There are a quite a few tools out there for assembling reveal.js decks in different ways, from different input. I think it's better if reveal.js remains more or less agnostic as to how the presentation HTML is built.

PR #2312, opened 2019-01-30
impove randomness

albert-ziegler:
I've noticed `Math.random()` used to generate a secret. However, its values are not safe against prediction, so if I understand the context correctly, it's probably preferable to use `Window.crypto` instead.

Full disclosure: I noticed this using [LGTM.com](https://lgtm.com/projects/g/hakimel/reveal.js/alerts/?mode=list&result_filter=0e00f5fa7b93849767c1677dc71d9400f85c6a54), a code analysis platform which I also work for.

### DIFF
--- a/plugin/multiplex/index.js
+++ b/plugin/multiplex/index.js
@@ -45,7 +45,7 @@ app.get("/", function(req, res) {
 app.get("/token", function(req,res) {
 	var ts = new Date().getTime();
 	var rand = new Uint32Array(1);
-	window.crypto.getRandomValues(array)
+	window.crypto.getRandomValues(rand)
 	var secret = ts.toString() + rand.toString();
 	res.send({secret: secret, socketId: createHash(secret)});
 });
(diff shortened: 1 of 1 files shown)


plugin/multiplex/index.js
```diff
@@ -44,7 +44,8 @@ app.get("/", function(req, res) {
 
 app.get("/token", function(req,res) {
 	var ts = new Date().getTime();
-	var rand = Math.floor(Math.random()*9999999);
+	var rand = new Uint32Array(1);
+	window.crypto.getRandomValues(array)
```
lirantal: probably a typo @albert-ziegler? I think you meant to put `rand` in that function call

albert-ziegler: Indeed. Thanks for spotting, @lirantal!

hakimel:
Thanks!

hakimel:
Had to revert this, `window.crypto` doesn't exist in Node.js so this threw an exception that prevented the multiplex server from working.

PR #2315, opened 2019-01-31
allow custom slide numbering functions

dougalsutherland:
As is, slide numbers can only use one of four hardcoded formats, and there's no way in the API to change them (other than I suppose rewriting the `innerHTML` of the slide numbers box after reveal has rewritten itself in a `slidechanged` signal handler, but that's pretty gross).

For example, I use vertical slides for backup slides, which means that if I'm going through the presentation and don't use them then my slide numbers go `4`, `5.1`, `6`. I'd rather have it go `4`, `5`, `6`, and call the vertical slides below it `5/1`, `5/2`, etc.

This two-line change (most of the diff is indentation...) allows for custom schemes by setting `config.slideNumber`:
```javascript
slideNumber: function() {
    var idx = Reveal.getIndices();
    var value = [idx.h + 1];
    if (idx.v > 0) { // first slide in vertical stacks are just plain
        value.push('/');
        value.push(idx.v);
    }
    return value;
},
```

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -3257,40 +3257,48 @@
 	 *  "h/v":	horizontal / vertical slide number
 	 *    "c":	flattened slide number
 	 *  "c/t":	flattened slide number / total slides
+	 *
+	 * Alternatively, config.slideNumber can be a function returning a
+	 * three-element array with arguments to formatSlideNumber().
 	 */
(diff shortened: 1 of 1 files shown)


hakimel:
Makes sense to make that configurable, thanks!

PR #2326, opened 2019-02-19
Allow Navigation Mode setting to change the shortcuts table

### LINKED ISSUE #2313: Use navigation_mode in keyboard shortcuts table
Since `navigation_mode` changes the behavior of the navigation keys, keyboard shortcuts table should be updated to display the relevant actions regarding the value of `navigation mode`.

earboxer:
Close #2313

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -361,20 +361,8 @@
 			threshold: 40
 		},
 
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks for fixing that!

PR #2333, opened 2019-02-28
remove unused `isPrintingPDFFragments` function

dougalsutherland:
`isPrintingPDFFragments` was in the original PR of #1955, but when you changed it to do per-fragment by default in 3680f1ad10d1cbb4a48eb98673fa7018d1ab36e5 you left this now-unused helper function in. Not used and not in the API, so no reason to keep it. :)

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -1830,15 +1830,6 @@
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


hakimel:
Good catch, thanks!

PR #2337, opened 2019-03-02
honor slide number config in pdf mode

dougalsutherland:
Currently, `print-pdf` mode always uses `h.v` slide numbers. This change makes it support any format.

This required changing the API of the custom slide number function from #2315. Since that's only been in `dev` for less than a month, I imagine it's not widely used yet, and using the old style of function would only break `print-pdf` numbering and not regular numbers anyway.


hakimel:
This is great! Thanks for a thorough PR that covers all the cases.

PR #2338, opened 2019-03-03
add text effects for span grow and shrink fragments

### LINKED ISSUE #2126: Grow and shrink fragment styles not working on span within P
As discussed on [pandoc-discuss](https://groups.google.com/d/msg/pandoc-discuss/MtKe52maiGE/uGSIYMQDBwAJ), pandoc will generate HTML from this in which grow and shrink do not work.

https://reagle.org/joseph/talks/2014/exemplar.html#/fragment-styles-new

```
## Fragment styles new

::: {.incremental}

[current-visible]{.fragment .current-visible}

[highlight-red]{.fragment .highlight-red}

[highlight-green]{.fragment .highlight-green}

[highlight-blue]{.fragment .highlight-blue}

[fade-out]{.fragment .fade-out}

[grow]{.fragment .grow} <!-- https://github.com/hakimel/reveal.js/issues/2126 -->

[shrink]{.fragment .shrink} <!-- https://github.com/hakimel/reveal.js/issues/2126 -->

:::
```

lunchboxer:
Fixes #2126

Since inline elements don't get transform scale effects, if the grow or shrink fragment class is applied to a span, nothing happens currently. Changing the font-size instead gives the expected result.

### DIFF
--- a/css/reveal.css
+++ b/css/reveal.css
@@ -76,7 +76,7 @@ body {
     -webkit-transform: scale(0.7);
             transform: scale(0.7); }
 
-.reveal .slides section span.fragment.grow.visible {
+.reveal .slides section span.fragment.shrink.visible {
   font-size: 70%;
 }        
       
--- a/css/reveal.css
+++ b/css/reveal.css
@@ -65,6 +65,10 @@ body {
     -webkit-transform: scale(1.3);
             transform: scale(1.3); }
 
+.reveal .slides section span.fragment.grow.visible {
+  font-size: 130%;
+}         
+
 .reveal .slides section .fragment.shrink {
   opacity: 1;
   visibility: inherit; }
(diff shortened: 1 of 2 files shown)


css/reveal.css
```diff
@@ -65,13 +65,21 @@ body {
...
+}         
+
 .reveal .slides section .fragment.shrink {
   opacity: 1;
   visibility: inherit; }
   .reveal .slides section .fragment.shrink.visible {
     -webkit-transform: scale(0.7);
             transform: scale(0.7); }
 
+.reveal .slides section span.fragment.grow.visible {
```
vnpavanelli: This would be shrink and not grow

lunchboxer: Yes, thank you for pointing that out. It should be fixed now

Issue #2176, opened 2018-06-09
Zoom does not work on Linux


rugk:
On your example site https://revealjs.com/#/4 the zoom does not work. I do Alt+click, but it does not work in the demo site all.

Firefox 60

Koriit:
@rugk, what is your OS? On Linux

rugk:
Well…˝the slide said "Alt+click"… but yes, it works with Ctrl… thanks.

Just wonder why it does not use `Alt` there, too. It would totally be possible…

Koriit:
According to @linux-man in #1909, ALT does not work on Linux.

linux-man:
By defult, ALT+Click grabs the window from any spot of the window to easily move it around without having precisely aim for the titlebar like one would usually.

linux-man:
I think it's the default behavior on KDE, Cinnamon, Mate, XFCE, Unity and most of the smaller DE. Gnome changed to SUPER+Click.

linux-man:
You can change this behaviour on inicialize:

```
Reveal.initialize({
	// ...
	zoomKey: "alt",
	// ...
	width: "100%",
	height: "100%"
}); 
```
or after:

```
Reveal.configure({
	zoomKey: "alt"
});
```

rugk:
Yeah, I'd rather worry about the standards. I mean, at least, it should be properly documented…

Or could not you just bind to both keys, so whatever one you can use on the platform works?

hakimel:
fixed by @sophware in #2340

PR #2340, opened 2019-03-06
demo.html - Added Linux instructions for zoom

sophware:
Fixes [issue 2176](https://github.com/hakimel/reveal.js/issues/2176).

In Linux, zoom.js uses ctrl + click, instead of alt + click. This was missing in the relevant slide in the demo at made Linux users think zoom wasn't working. The slide language has been changed and now includes a note mentioning the Linux-specific way to zoom.


hakimel:
makes sense, thanks!

PR #2346, opened 2019-03-12
notes plugin: callRevealApi wants apply, not call

dougalsutherland:
The `callRevealApi()` function introduced in #2104, which takes `methodArguments` which is clearly supposed to be a list, was passing that list as the first argument to the given function. This didn't really matter, because it's currently only ever called with `[]` and for functions that don't actually take any arguments, but it broke things in my #2337 that introduces an argument to `getSlidesPastCount`.


hakimel:
Good catch. Thanks!

PR #2350, opened 2019-03-15
Fixed not bing able to follow same inner links twice with history:false

rparree:
Scenario:

- history set to false in Reveal’s config
- follow anchor to `/#someid`
- navigate to other sections
- follow anchor to same section `/#someid`

This does not work as the hash has not changed.

This fix removes the hash from the URL when navigating


hakimel:
Thanks for the patch! I've merged this but hade to make some changes to make it play well with some newer code in the dev branch.

rparree:
For some reason when i was testing this it was working, but i am now using it in a production piece and it does not work. Changing the hash triggers a slide change.

Using the following code instead seems to do the trick:

```javascript
history.pushState("", document.title, window.location.pathname + window.location.search);
```

See also https://github.com/rparree/reveal.js/blob/hash-problem/js/reveal.js

Shall i make another pull request?

hakimel:
Hm, I tested this again with the 3.8.0 release but from what I can tell your original fix is working. Any chance you can include a sample presentation where the same link can't be followed twice?

rparree:
I tested it  with 3.8.0 and it works, so no need for another pull request. I have deleted my form meanwhile.

tx.,

PR #2353, opened 2019-03-18
Add data-preload attribute for iframes

### LINKED ISSUE #2352: Why shouldn't lazy-loaded iframes preload?
Looking at [this line](https://github.com/hakimel/reveal.js/blob/0b3e7839ebf4ed8b6c180aca0abafa28c67aee6d/js/reveal.js#L3721), if an iframe is contained in a fragment and has the `data-src` attribute, it won't be loaded until its containing fragment is visible. That means that when the fragment becomes visible, the page will begin visibly loading onscreen in the middle of the presentation. What was the intention behind this behavior? Why shouldn't iframes in fragments be preloaded per the viewDistance and simply hidden like non-fragment iframes? Would you be open to changing this behavior or adding an option for it?

FWIW I've worked around this behavior by simply not lazy-loading iframes.

maxrothman:
Allows lazy-loaded (i.e. data-src) frames to be preloaded when they come
within the viewDistance, rather than once they're visible.

Fixes #2352

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,13 @@ Reveal.initialize({
 	// - false: No media will autoplay, regardless of individual setting
 	autoPlayMedia: null,
 
+  // Global override for preloading lazy-loaded iframes
+	// - null: Iframes with data-src AND data-preload will be loaded when within
+	//   the viewDistance, iframes with only data-src will be loaded when visible
+	// - true: All iframes with data-src will be loaded when within the viewDistance
+	// - false: All iframes with data-src will be loaded only when visible
+	preload: null,
+
 	// Number of milliseconds between automatically proceeding to the
 	// next slide, disabled when set to 0, this value can be overwritten
 	// by using a data-autoslide attribute on your slides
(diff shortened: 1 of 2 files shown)


hakimel:
Thanks for the PR! The combination of a configured default + attribute overrides looks great. A couple of notes:
- The diff on `reveal.js` doesn't actually include the "preload" config value described in the readme. Maybe a commit is missing?
- I suggest renaming the `preload` config value to `preloadIframes` since it only affects iframes.
- Please submit this PR towards the dev branch. That's where the latest code lives and PRs are merged into before a new reveal.js version is released.

maxrothman:
> The diff on reveal.js doesn't actually include the "preload" config value described in the readme. Maybe a commit is missing?

Whoops, you're absolutely right! Thank goodness for code reviews.

> I suggest renaming the preload config value to preloadIframes since it only affects iframes.

So it'd be `data-preload` and `preloadIframes`, like `data-autoplay` and `autoPlayMedia`? That makes sense.

maxrothman:
Closing because this is against the wrong branch

PR #2364, opened 2019-04-06
Bugfix: Remove calls to missing modules from gruntfile

richardwestenra:
When running the app, Grunt throws the following warnings:
```
>> Local Npm module "grunt-contrib-clean" not found. Is it installed?
>> Local Npm module "grunt-contrib-nodeunit" not found. Is it installed?
```
These packages aren't included in package.json, and the modules don't appear to be used for anything, so I think they can just be removed.

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -162,9 +162,6 @@ module.exports = grunt => {
 
 	});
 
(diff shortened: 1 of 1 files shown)


hakimel:
Oops, those were accidentally left in. Thanks for the PR!

PR #2375, opened 2019-04-20
Switch to MathJax CommonHTML renderer

mbotsch:
# What

This pull request changes the MathJax renderer from HTML-CSS to CommonHTML in the math.js plugin. It also adjusts the notes plugin accordingly to have correct equations in speaker notes.

# Why

CommonHTML (or CHTML) is the default renderer of MathJax since version 2.6, the current version being 2.7.5. Reveal still uses the "old" HTML-CSS renderer. The rendered equations of the CHTML renderer, however, are more consistent/similar across different browsers, and in some cases even more correct. 

For instance, compare old/new for this equation:
$$\{\overrightarrow{S}\rightarrow\overrightarrow{L},\overleftarrow{S}\rightarrow\overleftarrow{L},\overleftarrow{L}\rightarrow\overleftarrow{L}\overrightarrow{S},\overleftarrow{L}\rightarrow\overleftarrow{S}\overrightarrow{L}\}$$

# How

- Select latest MathJax and CHTML configuration as default
- Disable MathJax' matchFontHeight, since this causes problems (for old and new renderer)
- Disable typesetting on slideChange, since it is no longer required
- When exporting a PDF: Use the new promise mechanism to ensure that math typesetting is finished before emitting the 'ready' signal. Otherwise, complex presentations with complex equations end up having some non-typeset equations in the PDF.
- In the speaker notes plugin, the main presentation send the location of the MathJax script to the speaker notes window, which loads the correct MathJax script and then has the correct math fonts.
- Disable AssistiveMML in MathJax configuration, since this setting erroneously duplicates all equations in speaker notes.

# Fixed Issues

Besides switching to the better (more consistent and more correct) math renderer, this pull request fixes the following open issues:
- 1924: Now works thanks to the new reset.css and the disabled matchFontHeight.
- 2105: This problem was caused by dynamically adjusting font height and by re-typesetting upon slide-change. Both can now safely be disabled.
- 1383: This problem was caused by re-typesetting math content on slide changes in overview mode. Both are disabled now.
- 1726: Disabling AssistiveMML removes duplicated math in notes.
- 2256: Solved by enforcing math is typeset in init() function through promises.

I'm using the CommonHTML renderer for math-heavy teaching for about a year now, and it works like a charm. I think it should be used simply because it has been the default choice of MathJax for quite some time. One particular benefit of the more consistent math typesetting is that manual annotation (using a pimped version of the chalkboard plugin) now match the precise location of math terms across different browsers. Let me know if I should prepare slides demonstrating before/after of the above issues.

Cheers
Mario

### DIFF
--- a/plugin/math/math.js
+++ b/plugin/math/math.js
@@ -7,17 +7,21 @@
 var RevealMath = window.RevealMath || (function(){
 
 	var options = Reveal.getConfig().math || {};
-	var mathjax = options.mathjax || 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.0/MathJax.js';
-	var config = options.config || 'TeX-AMS_HTML-full';
+	var mathjax = options.mathjax || 'https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js';
+	var config = options.config || 'TeX-AMS_CHTML-full';
 	var url = mathjax + '?config=' + config;
 
 	var defaultOptions = {
 		messageStyle: 'none',
+        extensions: ["TeX/AMSmath.js"],
 		tex2jax: {
 			inlineMath: [ [ '$', '$' ], [ '\\(', '\\)' ] ],
 			skipTags: [ 'script', 'noscript', 'style', 'textarea', 'pre' ]
 		},
-		skipStartupTypeset: true
+		skipStartupTypeset: true,
+        AssistiveMML: { disabled: true },
+        "CommonHTML": { matchFontHeight: false },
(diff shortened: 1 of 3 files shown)


mbotsch:
Is there anything that I can do to help processing this pull request?
If there is no interest in merging it, please also let me know.

bobbens:
Just want to comment that this worked for me fixing all issues I've been having with notes and mathematics.

mbotsch:
Thanks for confirming this!

BTW: I recently switched to the SVG renderer, since the/my current Chrome version does not render bold math fonts correctly. Let me know if you want details on this.

PR #2378, opened 2019-04-23
Fix plugin documentation regarding init-Promises

mbotsch:
Mention in the plugin section of README that reveal.js will wait for the Promise of a plugin's init() function only when the plugin is loaded non-async (see initPlugins()).

If a plugin is loaded as an async dependency, the wait-for-Promise mechanism in not applied. Reveal waits for the Promises of the non-sync plugins, then calls start(), which triggers the loading of the async plugins and dispatches 'ready'. This dispatching might happen before the async plugins are loaded and their init functions are done and their Promises are fullfilled.

Not sure that this behavior is intended. But I guess it should be documented. Maybe my take at improving the documentation is not perfect through.

BTW: The init-Promise mechanism is great!

Cheers
Mario

### DIFF
--- a/README.md
+++ b/README.md
@@ -1268,6 +1268,7 @@ Reveal.addEventListener( 'ready', () => console.log( 'Three seconds later...' ) 
 Reveal.initialize();
 ```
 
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks!

PR #2386, opened 2019-04-30
Add DejaVu *offline* fonts and a theme 'dejavu'

return42:
The Dejavu font has a high UTF coverage and is continuously developed.  WOFF2 is supported by all modern WEB browsers.  Using WOFF2 of Dejavu for *offline* scenarios and where CDNs are blocked, is obvious a good choice.

In this patch series the DejaVu fonts and a (offline) theme 'dejavu' which is using the (embedded) DejaVu fonts were added.


return42:
Project seems dead.

PR #2387, opened 2019-04-30
Handle lazy-loading of background iframes/videos

gcalmettes:
Currently, the background videos/iframes added via the `data-background-[video|iframe]` are lazy loaded correctly the first time when in the `viewDistance` of the current slide (and the `data-preload` attribute has been added for the specific case of the iframes). However, if we get out of the `viewDistance` distance, the background content lazy loaded is not unloaded like what happens for regular lazy loaded content on the main slides. This means that for example an iframe in the background content will still use the browser ressources even though it should have been unloaded.

This PR fixes this problem.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -3771,7 +3771,6 @@
 			}
 		} );
 
-
 		// Show the corresponding background element
 		var background = slide.slideBackgroundElement;
 		if( background ) {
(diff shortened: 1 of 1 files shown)


js/reveal.js
```diff
@@ -3876,13 +3873,12 @@
...
             iframe.style.maxWidth = '100%';
 
-            // If not on current slide, this iframe has been lazy loaded.
-            // Need to add the data-lazy-loaded attribute so Reveal will handle it
-            // with all of its other lazy loaded content
-            if ( !isCurrentSlide ) {
-              iframe.setAttribute('data-lazy-loaded', '');
-            }
-
+            // Add the lazy-loaded tag to ensure that iframe will be
```
gcalmettes: Note that adding the `data-lazy-loaded` tag to all the background iframes (and not only the ones with `lazy-loaded`) ensure that they will be unloaded (`src` to `data-src`) once out of the viewDistance.
If not, any iframe added on the background will never be unloaded throughout the presentation.

PR #2400, opened 2019-05-09
Notes: Introduce alternate pacing timer, based on total presentation time

fghaas:
The current pacing timer, introduced in 715cf0ba11db0f913cfd2edb56a2c66aed57b505 (#1564), operates on the assumption that there is a default amount of time to be allocated to each slide, and that individual slides can deviate from that default by specifying their own `data-timing` attribute.

This patch introduces an alternate pacing method: by specifying the `totalTime` configuration option, the presenter can set the total time available to present. The pacing timer will then continue to allocate the exact pacing time for slides that do have `data-timing` set, as before. However, rather than applying the `defaultTiming` constant to all others, it will

- Add up the time already allocated to slides via `data-timing`;
- subtract that from totalTime;
- divide the difference by the number of slides *without* `data-timing` set;
- apply the thus-calculated average to those slides.

`totalTime` has no default, and if both `defaultTiming` and `totalTime` are set, `totalTime wins`. This preserves backward compatibility: if a presenter has set `defaultTiming` and updates reveal.js, `totalTime` will be `null` and `defaultTiming` is still applied to all slides without a `data-timing` attribute. The presenter can then switch to the automatic calculation, if desired, by setting a value for `totalTime`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,6 +346,13 @@ Reveal.initialize({
 	// instead of using the defaultTiming value
 	totalTime: 0,
 
+	// Specify the minimum amount of time you want to allot to
+	// each slide, if using the totalTime calculation method.  If
+	// the automated time allocation causes slide pacing to fall
+	// below this threshold, then you will see an alert in the
+	// speaker notes window
+	minimumTimePerSlide: 0;
+
 	// Enable slide navigation via mouse wheel
 	mouseWheel: false,
 
--- a/README.md
+++ b/README.md
@@ -340,6 +340,12 @@ Reveal.initialize({
 	// speaker view
 	defaultTiming: 120,
 
+	// Specify the total time in seconds that is available to
+	// present.  If this is set to a nonzero value, the pacing
+	// timer will work out the time available for each slide,
+	// instead of using the defaultTiming value
+	totalTime: 0,
+
 	// Enable slide navigation via mouse wheel
 	mouseWheel: false,
 
@@ -1228,7 +1234,7 @@ The speaker notes window will also show:
 - Current wall-clock time
 - (Optionally) a pacing timer which indicates whether the current pace of the presentation is on track for the right timing (shown in green), and if not, whether the presenter should speed up (shown in red) or has the luxury of slowing down (blue).
 
-The pacing timer can be enabled by configuring by the `defaultTiming` parameter in the `Reveal` configuration block, which specifies the number of seconds per slide.  120 can be a reasonable rule of thumb.  Timings can also be given per slide `<section>` by setting the `data-timing` attribute.  Both values are in numbers of seconds.
+The pacing timer can be enabled by configuring the `defaultTiming` parameter in the `Reveal` configuration block, which specifies the number of seconds per slide.  120 can be a reasonable rule of thumb.  Alternatively, you can enable the timer by setting `totalTime`, which sets the total length of your presentation (also in seconds).  If both values are specified, `totalTime` wins and `defaultTiming` is ignored.  Regardless of the baseline timing method, timings can also be given per slide `<section>` by setting the `data-timing` attribute (again, in seconds).
 
 
 ## Server Side Speaker Notes
--- a/plugin/notes/notes.html
+++ b/plugin/notes/notes.html
@@ -540,6 +540,7 @@
 					callRevealApi( 'getSlidesAttributes', [], function ( slideAttributes ) {
 						callRevealApi( 'getConfig', [], function ( config ) {
 							var totalTime = config.totalTime;
+							var minTimePerSlide = config.minimumTimePerSlide || 0;
 							var defaultTiming = config.defaultTiming;
 							if ((defaultTiming == null) && (totalTime == null)) {
 								callback(null);
@@ -573,6 +574,11 @@
 								var timePerSlide = Math.round( remainingTime / remainingSlides, 0 )
 								// And now we replace every zero-value timing with that average
 								timings = timings.map( function(x) { return (x==0 ? timePerSlide : x) } );
+							}
+							var slidesUnderMinimum = timings.filter( function(x) { return (x < minTimePerSlide) } ).length
+							if ( slidesUnderMinimum ) {
+								message = "The pacing time for " + slidesUnderMinimum + " slide(s) is under the configured minimum of " + minTimePerSlide + " seconds. Check the data-timing attribute on individual slides, or consider increasing the totalTime or minimumTimePerSlide configuration options (or removing some slides).";
+								alert(message);
 							}
 							callback( timings );
 						} );
--- a/plugin/notes/notes.html
+++ b/plugin/notes/notes.html
@@ -539,12 +539,16 @@
 
 					callRevealApi( 'getSlidesAttributes', [], function ( slideAttributes ) {
 						callRevealApi( 'getConfig', [], function ( config ) {
+							var totalTime = config.totalTime;
 							var defaultTiming = config.defaultTiming;
-							if (defaultTiming == null) {
+							if ((defaultTiming == null) && (totalTime == null)) {
 								callback(null);
 								return;
 							}
-
+							// Setting totalTime overrides defaultTiming
+							if (totalTime) {
+								defaultTiming = 0;
+							}
 							var timings = [];
 							for ( var i in slideAttributes ) {
 								var slide = slideAttributes[ i ];
@@ -559,7 +563,17 @@
 								}
 								timings.push(timing);
 							}
-
+							if ( totalTime ) {
+								// After we've allocated time to individual slides, we summarize it and
+								// subtract it from the total time
+								var remainingTime = totalTime - timings.reduce( function(a, b) { return a + b; }, 0 );
+								// The remaining time is divided by the number of slides that have 0 seconds
+								// allocated at the moment, giving the average time-per-slide on the remaining slides
+								var remainingSlides = (timings.filter( function(x) { return x == 0 }) ).length
+								var timePerSlide = Math.round( remainingTime / remainingSlides, 0 )
+								// And now we replace every zero-value timing with that average
+								timings = timings.map( function(x) { return (x==0 ? timePerSlide : x) } );
+							}
 							callback( timings );
 						} );
 					} );


fghaas:
@aspiers, since you're the original author of this feature, would you be so kind to take a look and see if you think this is useful?

aspiers:
Brilliant idea, I love it!

Couple of thoughts:

- It would still be backwards-compatible even if `totalTime` overrode `defaultTiming` when both are set, because older decks would not have `totalTime` set and therefore not change in behaviour when upgraded to newer `reveal.js`.  But the precedence you propose is also fine at first thought.

- We could go even further (in the future, not in this PR), and dynamically and proportionally adapt timings as the presentation progresses, in order to "home in" on the desired completion time?  For example if there are 25 minutes worth of slides left, but you're running behind  and only have 20 minutes left, it could dynamically shave 20% off each slide in order to try to help you get back on track.  OTOH ...
    - This could be confusing.
    - You might or might not want it to add time if you get *ahead* of your plans.
    - If you temporarily jumped backwards to quickly show an older slide (e.g. if someone in the audience asks a question about one), then the timing would temporarily look very scary until you returned to your original location.  This could be mitigated by keeping track of the furthest on slide shown so far, and dynamically calculating timings based on that rather than on the currently shown slide.

fghaas:
@aspiers Thanks for looking into this!

> It would still be backwards-compatible even if totalTime overrode defaultTiming when both are set, because older decks would not have totalTime set and therefore not change in behaviour when upgraded to newer reveal.js.

That's a good point.

> We could go even further (in the future, not in this PR), and dynamically and proportionally adapt timings as the presentation progresses, in order to "home in" on the desired completion time?

Right, in the future; I'll leave the elliptic-curve implementation involving lambda calculus to you. ;)

I do have a couple of lower-hanging fruit though that I might still want to fix, after sleeping on this for a night:

- The `totalTime` approach can presently result in an arbitrarily short calculated `timing`, including one that is negative. That is clearly not good. I'm thinking about adding another config option that sets the minimum time-per-slide, in seconds,¹ and uses `alert()` to notify the speaker if any slide is below that.
- When setting `totalTime`, having an elapsed time counter that counts *forwards* is a bit silly; it should probably be countdown timer that count backwards instead. But I don't know what's a good and intuitive behaviour for clicking the timer, then. When you introduced `defaultTiming`, you made the fine choice of altering the behavior from "click means start over" to "click means right on track," and I wonder what's the good thing to do in the `totalTime` case. Does click mean "OK so it's total time from *now*" (meaning the end time is later), or does it mean "on track" (meaning the end time is also later, but not by as much)?
 * * *
¹ What's a good default? Probably 0, because people might use Lessig style slide decks and go really really fast.

README.md
```diff
@@ -1228,7 +1234,7 @@ The speaker notes window will also show:
 - Current wall-clock time
 - (Optionally) a pacing timer which indicates whether the current pace of the presentation is on track for the right timing (shown in green), and if not, whether the presenter should speed up (shown in red) or has the luxury of slowing down (blue).
 
-The pacing timer can be enabled by configuring by the `defaultTiming` parameter in the `Reveal` configuration block, which specifies the number of seconds per slide.  120 can be a reasonable rule of thumb.  Timings can also be given per slide `<section>` by setting the `data-timing` attribute.  Both values are in numbers of seconds.
+The pacing timer be enabled by configuring by the `defaultTiming` parameter in the `Reveal` configuration block, which specifies the number of seconds per slide.  120 can be a reasonable rule of thumb.  Alternatively, you can enable the timer by setting `totalTime`, which sets the total length of your presentation (also in seconds).  If both values are specified, `defaultTiming` wins and `totalTime` is ignored.  Regardless of the baseline timing method, timings can also be given per slide `<section>` by setting the `data-timing` attribute (again, in seconds).
```
fghaas: Note to self: missing "can", duplicate "by". (Will fix)

plugin/notes/notes.html
```diff
@@ -559,7 +562,17 @@ <h4 class="label">Notes</h4>
 								}
 								timings.push(timing);
 							}
-
+							if ( totalTime ) {
+								// After we've allocated time to individual slides, we summarize it and
+								// subtract it from the total time
+								var remainingTime = totalTime - timings.reduce( function(a, b) { return a + b; }, 0 );
```
fghaas: Note to self: this can be negative, resulting in timePerSlide being negative for all slides as well. Not good. Will fix.

README.md
fghaas: Fixed in 9cd3cfbcd83fcb522cd329a3fa6ec2134b86902a.

plugin/notes/notes.html
fghaas: First stab at a fix for this is in 5bf78d2967b1706dbfda06d27512c13fc4f12048.

fghaas:
@aspiers: patch updated. 

* Followed your suggestion about coniguration defaults.
* Implemented a very weak "enforcement" of minimum time-per-slide. Currently uses `console.warn()`, which the presenter will probably not look at. Considered using `alert()` but that's probably too invasive, and whether or not it makes sense to include this information in the pacing timer window I'm not sure. Thoughts?

aspiers:
[**@fghaas**](https://github.com/fghaas) commented on [May 10, 2019 8:52 AM](https://github.com/hakimel/reveal.js/pull/2400#issuecomment-491194940 "2019-05-10T07:52:18Z - Replied by Github Reply Comments"):
> I do have a couple of lower-hanging fruit though that I might still want to fix, after sleeping on this for a night:
> 
> *   The `totalTime` approach can presently result in an arbitrarily short calculated `timing`, including one that is negative. That is clearly not good. I'm thinking about adding another config option that sets the minimum time-per-slide, in seconds,¹ and using `alert()` to notify the speaker if any slide is below that.

Makes sense.

> *   When setting `totalTime`, having an elapsed time counter that counts _forwards_ is a bit silly; it should probably be a countdown timer counting backwards instead. But I don't know what's a good and intuitive behaviour for _clicking_ the timer, then. When you introduced `defaultTiming`, you made the fine choice of altering the behavior from "click means start over" to "click means right on track," and I wonder what's the good thing to do in the `totalTime` case. Does click mean "OK so it's total time from _now_" (meaning the end time is later), or does it mean "on track" (meaning the end time is also later, but not by as much)?

IIUC the former feels like a regression in behaviour, and would make decreasing amounts of sense the further through the deck you are when you click.  If `totalTime` is set to `30` and you click with only two slides to go neither of which have `data-timing` set, that's going to allocate 15 minutes per slide, right?  Which doesn't provide any obvious value.  OTOH if you were originally given 30 minutes for the slot but you use that all up with two slides remaining, and the event organiser at the back of the room says "don't worry, you can have 5 more minutes", then the latter approach would mean that clicking gives you either 2 or 2.5 minutes per slide depending on how you calculate it, and both are much more useful.  Or did I misunderstand something?

> ¹ What's a good default? Probably 0, because people might use Lessig style slide decks and go really really fast.

Well, 0 seconds per slide is *infinitely* fast, so that doesn't make much sense to me ;-)  I'd suggest 2 seconds per slide as a minimum, since that gives a speaker just about enough time to load a slide (including heavy images etc.) and make a split-second decision to skip it.
[**@fghaas**](https://github.com/fghaas) commented on [May 12, 2019 12:37 PM](https://github.com/hakimel/reveal.js/pull/2400#issuecomment-491587906 "2019-05-12T11:37:40Z - Replied by Github Reply Comments"):

> [@aspiers](https://github.com/aspiers): patch updated.
> 
> *   Followed your suggestion about configuration defaults.

I thought I only made an observation rather than any helpful suggestion, but OK :-)

> *   Implemented a very weak "enforcement" of minimum time-per-slide. Currently uses `console.warn()`, which the presenter will probably not look at.

Yeah IMHO that's too hidden.

> Considered using `alert()` but that's probably too invasive

If it always appears as soon as the deck loads then shouldn't this be fine, assuming that any sane presenter will try loading their deck at least *once* before the actual live presentation? 

But even if that still worries you for some reason:

> and whether or not it makes sense to include this information in the pacing timer window I'm not sure. Thoughts?

Yes, I think that could be an OK place for it, maybe even in addition to the `alert()`.  I think there could be presenters who forget to open the speaker notes window (or at least pay close attention to it) until the actual live presentation.

Either way, the important thing is to phrase the warning in a way which makes it clear how to fix the issue.

fghaas:
@aspiers Having run through some talk rehearsals using this, I have come to the conclusion that the click behaviour needs no changes at all. It's perfectly useful as-is, *particularly* during practice runs.

> If it always appears as soon as the deck loads then shouldn't this be fine, assuming that any sane presenter will try loading their deck at least once before the actual live presentation?

No it wouldn't appear right when the deck loads but when you open the speaker view — but your comment still applies; any sane speaker would rehearse with notes at least once before presenting.

plugin/notes/notes.html
```diff
@@ -574,6 +575,11 @@ <h4 class="label">Notes</h4>
 								// And now we replace every zero-value timing with that average
 								timings = timings.map( function(x) { return (x==0 ? timePerSlide : x) } );
 							}
+							var slidesUnderMinimum = timings.filter( function(x) { return (x < minTimePerSlide) } ).length
+							if ( slidesUnderMinimum ) {
+								message = "The calculated pacing time for " + slidesUnderMinimum + " slide(s) is under the configured minimum of " + minTimePerSlide + " seconds. Check the data-timing attribute on individual slides, or consider increasing the totalTime or minimumTimePerSlide configuration options (or removing some slides).";
```
fghaas: > Either way, the important thing is to phrase the warning in a way which makes it clear how to fix the issue.

@aspiers: Would you think that this is clear enough for the alert message?

aspiers: I suggest this small tweak:

```suggestion
								message = "The calculated pacing time for " + slidesUnderMinimum + " slide(s) without the data-timing attribute is under the configured minimum of " + minTimePerSlide + " seconds. Check the data-timing attribute on other slides, or consider increasing the totalTime or minimumTimePerSlide configuration options (or removing some slides).";
```

aspiers:
[**@fghaas**](https://github.com/fghaas) commented on [May 12, 2019 3:54 PM](https://github.com/hakimel/reveal.js/pull/2400#issuecomment-491602709 "2019-05-12T14:54:32Z - Replied by Github Reply Comments"):
> [@aspiers](https://github.com/aspiers) Having run through some talk rehearsals using this, I have come to the conclusion that the click behaviour needs no changes at all. It's perfectly useful as-is, _particularly_ during practice runs.

Cool :-)

> > If it always appears as soon as the deck loads then shouldn't this be fine, assuming that any sane presenter will try loading their deck at least once before the actual live presentation?
> 
> No it wouldn't appear right when the deck loads but when you open the speaker view — but your comment still applies; any sane speaker would rehearse with notes at least once before presenting.

Ah OK, that sounds fine: if they don't rehearse with notes then I guess they aren't planning to use them at all, in which case timing is irrelevant anyway.

plugin/notes/notes.html
fghaas: That tweak is at odds with what the code does, though. :)

The preceding test checks _all_ slides for whether they're under the minimum, including the ones that have `data-timing` set. I was on the fence about what to do here, but figured that if you do set a nonzero minimum time per slide, then you'd also want to be warned when a slide _with_ `data-timing` set the pacing time (presumably unintentionally) below the intended minimum. Say your minimum is 5, and you meant to set `data-timing` on a slide to 30 but typoed it to 3, I think it's not unreasonable to assume you'd like to be warned about that.

fghaas:
> Ah OK, that sounds fine: if they don't rehearse with notes then I guess they aren't planning to use them at all, in which case timing is irrelevant anyway.

Completely agree.

plugin/notes/notes.html
aspiers: Ah, good point. How about making it more explicit so that it explains whether slides with the attribute are below the limit, or slides without are, or some of both?

fghaas: In the interest of future contributors/maintainers, I'd prefer to keep this change minimal, and forgo that idea.

I'm really not trying to sound lazy here, but unless I'm mistaken then for your suggestion to work it'd take a separate counter for `data-timing` enabled slides. And I think that would just make this bit more convoluted (read: less maintainable), for little tangible benefit to the user. If a presenter is using the pacing timer — which you can only enable by editing the configuration —, chances are that they're well capable of running a recursive grep for the strings `dataTiming`, `totalTime`, and `minimumTimePerSlide`.

This stance might sound ridiculous for a tiny change like that, but those changes do add up — and if I can make Hakim's life just a wee bit easier without harming UX, I'll take that opportunity. :)

aspiers: Sounds good to me! You are doing the right thing by favouring simplicity; I have a tendency to over-engineer things 😜

fghaas: Excellent, glad we agree. Would you say that I can go ahead and propose the PR in its current state? If so, then please resolve this conversation and add your +1 on the review — otherwise, please send me any other nits you want me to pinch. ;)

aspiers: Seems I don't have enough permissions here for a Resolve Conversation button, but I've approved the PR.

fghaas: Huh, interesting. OK then I'll go ahead and do that. :)

fghaas:
@hakimel Courtesy ping here, in case you want to consider merging this. Thank you!

hakimel:
This is great! Will test it out and merge soon.

fghaas:
@hakimel, I added one more force-push that removes the word "calculated" from the pop-up alert, because the alert does (intentionally) also complain if any `data-timing` attribute on a slide is under `minimumTimePerSlide`:

```diff
diff --git a/plugin/notes/notes.html b/plugin/notes/notes.html
index ac6b11d..fdb81ad 100644
--- a/plugin/notes/notes.html
+++ b/plugin/notes/notes.html
@@ -577,7 +577,7 @@
                                                        }
                                                        var slidesUnderMinimum = timings.filter( function(x) { return (x < minTimePerSlide) } ).length
                                                        if ( slidesUnderMinimum ) {
-                                                               message = "The calculated pacing time for " + slidesUnderMinimum + " slide(s) is under the configured minimum of " + minTimePerSlide + " seconds. Check the data-timing attribute on individual slides, or consider increasing the totalTime or minimumTimePerSlide configuration options (or removing some slides).";
+                                                               message = "The pacing time for " + slidesUnderMinimum + " slide(s) is under the configured minimum of " + minTimePerSlide + " seconds. Check the data-timing attribute on individual slides, or consider increasing the totalTime or minimumTimePerSlide configuration options (or removing some slides).";
                                                                alert(message);
                                                        }
                                                        callback( timings );
```

hakimel:
Merged! Thanks for a great PR @fghaas.

I recently added support for total time to the slides.com speaker view as well. My goal is to bring over some of that design to the reveal.js speaker view. More specifically;
- Count-up timer turns into a countdown if a total time is provided.
- Showing presentation progress vs remaining time as visual bars makes it easy to see where you're at, at a glance.
- Timer can be paused/reset.

fghaas:
@hakimel & @aspiers, thanks to both of you!

PR #2410, opened 2019-05-18
Fix arrow down navigation on touch devices

NoriSte:
Some users reported me that the arrow down doesn't work on touch devices. They found that the arrow down brings them to the last slide sometimes.

After some investigations, I and @Ir0nsides have realized that it's not a bug per se, it's just a usability problem. There is a (too much) subtle space between the arrows and the progress bar that, on a touch device, make the user tap on the progress bar instead of on the arrow.

What we have done: we augmented the sensible area of the arrow down to avoid this kind of problems. It, obviously, make impossible to jump straight to the last slide when touching the progress bar (if the down arrow is visible), but try the current version on a touch device to realize how much is the current behavior annoying.

[Open Source Saturday](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/)
I'd like to spread some love for the [Open Source Saturday](https://www.meetup.com/it-IT/Open-Source-Saturday-Milano/) meetups, a series of events where the only goal is to contribute to OSS ❤️

### DIFF
--- a/css/reveal.css
+++ b/css/reveal.css
@@ -227,7 +227,7 @@ body {
   bottom: 12px;
   right: 12px;
   left: auto;
-  z-index: 1;
+  z-index: 11;
   color: #000;
   pointer-events: none;
   font-size: 10px; }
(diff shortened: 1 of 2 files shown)


hakimel:
Thanks for the PR! I ran into this issue the other day too and I'm sure many others have as well.

denehyg:
Hey @hakimel, I submitted PR #2142 a while back to implement a passive option for the progress bar for the same issue. It’s not just the down arrow that’s the issue. If there’s interactive content near the bottom, eg full slide data viz, it can be triggered from anywhere along the progress bar. A passive option would avoid this.

NoriSte:
I think that @denehyg should be merged too, it solves a more generic problem 👍

PR #2416, opened 2019-05-29
Linear Navigation: Fix to work on mobile, fix missing arrows

### LINKED ISSUE #2380: linear navigtationMode does not work in mobile
When I change the navigationMode to linear, it works fine for desktop view but when swiping through mobile it still uses the default navigation. Is there a way to configure the navigationMode to work for both desktop and mobile in the same way?

earboxer:
Close #2380 

Also close an issue with the arrows not appearing if the last or first slide are vertical stacks.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -3951,7 +3951,7 @@
 			}
 		}
 
-		if( config.navigationMode === "linear" ) {
+		if( config.navigationMode === 'linear' ) {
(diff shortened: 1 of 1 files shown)


hakimel:
Merged, thanks!

I had to remove the change you made to the control arrows. With that change, the navigation arrows were broken in a presentation that consists only of vertical slides.

PR #2419, opened 2019-05-29
Allow Reveal to be initialized multiple times

rogatty:
Since https://github.com/hakimel/reveal.js/commit/6bfa48a667fe91bb1de1b680769aeb7260af3a49 there is no way to initialize Reveal multiple times. We have an SPA where there are many slideshows and the Reveal DOM is created and destroyed multiple times without page reload. I added an option that allows to avoid the default behaviour of allowing initialization only once.

### DIFF
--- a/README.md
+++ b/README.md
@@ -351,7 +351,7 @@ Reveal.initialize({
 	// the automated time allocation causes slide pacing to fall
 	// below this threshold, then you will see an alert in the
 	// speaker notes window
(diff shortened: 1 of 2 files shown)


rogatty:
Actually, it wasn't enough for us and I used this instead to reload the Reveal instance:
```
delete require.cache[require.resolve('vendor/reveal/reveal')];
const Reveal = require('vendor/reveal/reveal');
```

PR #2421, opened 2019-05-29
* SCSS: Pass colors as CSS Variables

earboxer:
Note: CSS needs to be compiled when this is released.

This allows presenters to easily use the colors from [CSS custom properties](https://caniuse.com/#feat=css-variables), like so

```html
<section>
<span width="100px" height="200px" style="background-color:var(--link-color); color:var(--background-color);">Odd scheme</span> for contrast!
</section>
```


earboxer:
Closing in favor of #2521

Issue #2418, opened 2019-05-29
Fullscreen button


jefferyyuan:
- For users who are not familiar with Reveal.js, they don't know the shortcut 'f" to switch to full screen.
- It would be great if there is a button like in <img src='https://png.pngtree.com/svg/20160707/3a6bf2578b.svg' height="20" width="20" />, so a new user can easily figure out how to switch to full screen.
- or maybe double click to switch to full screen.
- The full-screen button would especially help we embed and share the slide using an iframe. The users are usually not familiar with revealjs, and the iframe size is usually limited.

Thanks

NitinMadhukar:
can i solve this

denehyg:
Have a look at https://github.com/denehyg/reveal.js-toolbar - it will give you a full screen button

jefferyyuan:
Thanks, @denehyg , Just found that yesterday, it's exactly what I am looking for.
- Thanks and close the issue now.

PR #2422, opened 2019-05-30
Added Fullscreen button #2418 

NitinMadhukar:


earboxer:
https://github.com/hakimel/reveal.js/blob/master/CONTRIBUTING.md#plugins
> Please do not submit plugins as pull requests. They should be maintained in their own separate repository. More information here: https://github.com/hakimel/reveal.js/wiki/Plugin-Guidelines

hakimel:
Thanks for the PR! As was pointed out in @earboxer's comment, this plugin should love in a separate repo to make it easier to maintain.

PR #2426, opened 2019-06-01
Changed http to https

sm-raiyyan:


sm-raiyyan:
Changed http to https

Issue #2175, opened 2018-06-09
Is it possible to stay on current slide after reload during development?


lichtner:
While I am developing a presentation, I would like to stay on the current working slide. But after saving changes page is reloaded and presentation is moved to the first slide. It is very annoying. Is there any way to force reveal.js to stay on the current working slide during development? Thanks.

rugk:
Just set `history` to true. Then it should show it to you and reload to the same slide via `#/slide` number.

lichtner:
Excellent, you saved my life :smile:

melaniewalsh:
You just saved mine as well! Thank you @rugk!

hakimel:
There's a new `hash` config option that you use to address this as well. `history: true` means each slide change is pushed to browser history. `hash: true` gives you unique URLs for each slide without pushing to the browser history.

I've updated the default index.html to include `hash: true`.

PR #2434, opened 2019-06-07
set history to true by default to improve developer experience.

veggiemonk:
It took a while to understand why I had to start from the beginning for every change. Searching for that is not easy either.

See issue #2175

`
While I am developing a presentation, I would like to stay on the current working slide. But after saving changes page is reloaded and presentation is moved to the first slide. It is very annoying. Is there any way to force reveal.js to stay on the current working slide during development? Thanks.
`

possible improvement: add a npm command such as `npm dev` which will automatically set history to true without writing it inside the reveal.js Init function.

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -162,9 +162,6 @@ module.exports = grunt => {
 
 	});
 
-	grunt.loadNpmTasks('grunt-contrib-clean');
-	grunt.loadNpmTasks('grunt-contrib-nodeunit');
-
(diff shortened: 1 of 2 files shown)


hakimel:
Thanks for the PR. 3.8.0 added a new `hash` config option. `hash: true` gives you unique URLs without pushing each slide change to the browser history. 

This option was already set in the demo.html file but I've added it to the index.html template now too.

PR #2442, opened 2019-06-22
Tweak to MathJax example

pius:
Fixed a missing comma and used an actual macro so that the MathJax example runs right out of the box upon copy/paste.


hakimel:
🙌

PR #2450, opened 2019-07-04
Fix light/dark bakground detection for backgrounds specified with hsl()

tlsa:
Adds support for HSL(A) colors to the color parsing code.

Previously with the default theme, a light HSL-specified background
did not change the text color to dark, making the text unreadable.

You can see the problem that this pull request fixes with the following slide, and the default theme:

```html
<section data-background-color="hsl(0, 30%, 100%)">Slide 1</section>
```

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -1824,6 +1824,77 @@
 	}
 
 	/**
+	 * Converts HSL color value to RGB.
+	 *
+	 * @param {h} hue (0 <= h <= 360)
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks for the PR!

From some quick testing in Chrome/FF I think `getComputedStyle(...).backgroundColor` is always in rgb/rgba format, even if the color was specified in hsl. Maybe instead of adding a hsl parser, we could read the rgb color from the element? Will need to experiment a bit more.

PR #2458, opened 2019-07-16
Fix errors with audit log fix

zolotyh:
Just a result of `npm audit fix`


hakimel:
Thanks for the PR but a fix for this was merged from another PR recently.

PR #2465, opened 2019-08-14
Added a reset function for when the dom drastically changes.

adunndevster:
Hi! I hope this is helpful for others as it is for me in my project! I needed a way to reset reveal, so that I could run multiple different presentations in the same browser instance without a hard refresh. I'm using Vue, but this should be helpful for other similar frameworks.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -382,6 +382,52 @@
 		registeredKeyBindings = {};
 
 	/**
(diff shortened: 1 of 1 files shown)


hakimel:
Hey. We already provide a `Reveal.sync()` method which you can call anytime there are changes to the presentation DOM. This should reset all of the internals based on what is currently in the DOM.

Let me know if you encounter any limitations with the sync method though.

PR #2488, opened 2019-09-28
Use npm ci instead of npm install

Samsinite:
Since the repository has a `package-lock.json`, recommend use of [npm ci](https://docs.npmjs.com/cli/ci.html) to provide reliabile and consistent installs.


hakimel:
Thanks for the suggestion but I prefer sticking to `npm install`. People know what it is, and `npm ci` seems mostly geared at CI/automated environments.

PR #2513, opened 2019-10-18
Add mobileViewDistance config key

TuurDutoit:
We use Reveal.js in a project where we only ever want to show 1 slide at a time, so we set `viewDistance: 1`. On mobile devices however, this setting is overriden to `viewDistance: 2` which breaks our application.
This PR adds a `mobileViewDistance` key which defaults to 2, as it was before, but can now be configured at will.


hakimel:
Looks great, thanks!

Issue #2064, opened 2018-01-05
data-noescape and data-trim cannot be added to code blocks in markdown


languitar:
I am unable to add marks to parts of highlighted source code in markdown slides.

One the one hand, the missing attributes `data-noescape` and possibly `data-trim` cannot be added to the code blocks from markdown using the `<!-- .element...` syntax.
On the other hand, inline html doesn't work correctly in case the code fragment contains newlines. In that case, the markdown parser breaks and separates the code fragment into multiple ones.

denisrosca:
@languitar Did you manage to somehow get around this issue? I'm having the same problem. 
It seems that empty attributes are completely ignored...

languitar:
I haven't found a solution so far.

Bagira80:
Following are some observations and partial solutions, that sadly does not lead to a fix for this issue:

---

I realized, that boolean attributes (like `data-noescape` and `data-trim`, which do not have a value) are not considered by the `plugin/markdown/markdown.js` script, when parsing the `<!-- .element...` syntax.

I managed to patch that script, so that at least boolean `data-...` attributes are successfully parsed and added to the former html elements.

```diff
--- plugin/markdown/markdown.js
+++ plugin/markdown/markdown.js
@@ -307,7 +307,7 @@
 	function addAttributeInElement( node, elementTarget, separator ) {
 
 		var mardownClassesInElementsRegex = new RegExp( separator, 'mg' );
-		var mardownClassRegex = new RegExp( "([^\"= ]+?)=\"([^\"=]+?)\"", 'mg' );
+		var mardownClassRegex = new RegExp( "([^\"= ]+?)=\"([^\"=]+?)\"|(data-[^\"= ]+?)(?=[\" ])", 'mg' );
 		var nodeValue = node.nodeValue;
 		if( matches = mardownClassesInElementsRegex.exec( nodeValue ) ) {
 
@@ -315,7 +315,11 @@
 			nodeValue = nodeValue.substring( 0, matches.index ) + nodeValue.substring( mardownClassesInElementsRegex.lastIndex );
 			node.nodeValue = nodeValue;
 			while( matchesClass = mardownClassRegex.exec( classes ) ) {
-				elementTarget.setAttribute( matchesClass[1], matchesClass[2] );
+				if( matchesClass[2] ) {
+					elementTarget.setAttribute( matchesClass[1], matchesClass[2] );
+				} else {
+					elementTarget.setAttribute( matchesClass[3], "" );
+				}
[...]

PR #2515, opened 2019-10-23
Markdown: Boolean (data-)attributes will now be added to elements, too.

### LINKED ISSUE #2038: failed to add element/slide attribute which includes equal signs
If the value of an attribute includes equal signs, then this attribute is ignored.

One example of the use case is setting `data-background-iframe` to a URL with a query string:

```html
<!-- .slide: data-background-iframe="https://www.youtube.com/embed/XXXX?rel=0" data-background-interactive -->
```

The reason of this seems to be in https://github.com/hakimel/reveal.js/blob/master/plugin/markdown/markdown.js#L285 , which requires the value of an attribute doesn't include `"` or `=`.

Bagira80:
When converting Markdown to HTML, boolean attributes (whose names start with `data-...`) will also be added to the resulting HTML elements, when attached via the `<!-- .element: ...` syntax.

Additionally, values of non-boolean attributes may now contain `=`. (fixes #2038)

### DIFF
--- a/plugin/markdown/markdown.js
+++ b/plugin/markdown/markdown.js
@@ -307,7 +307,7 @@
 	function addAttributeInElement( node, elementTarget, separator ) {
 
 		var mardownClassesInElementsRegex = new RegExp( separator, 'mg' );
-		var mardownClassRegex = new RegExp( "([^\"= ]+?)=\"([^\"=]+?)\"", 'mg' );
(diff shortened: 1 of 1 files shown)


Bagira80:
Note: This pull-request slightly touches #2064, but sadly does not fix it, too, because that problem is harder to solve and seems to involve hacking into `marked.js`.

hakimel:
Great fix, thanks!

Xanderthepoison:
'E-merging' technology

Issue #2524, opened 2019-11-06
Make Reveal more bundler-friendly


quilicicf:
## What am I trying to do

I'm trying to use a bundler (parceljs in my case) to generate a self-sufficient HTML file containing a Reveal.js presentation. 

This would have a lot of pros:
* generate a presentation in a universal format 
* the presentation would be tiny (the bundler can minify, compress etc...) and even tinier if zipped
* it would be easy to handle and pass around in an email or other means because it would be in a single file

## How is my presentation structured

Reveal.js is installed in my project via `npm`.

I have an `index.html` file containing my slides that imports a JS file. 

The JS file imports the CSS, plugins etc... and initializes Reveal.js.

> I can provide some code if that helps.

## The issues I am facing

Reveal.js is not currently bundler-friendly for a couple of reasons described below.

### Plugins

Reveal.js assumes the plugins are served over HTTP and it loads them from their source by injecting a `link` element with `src='relativeUrlSubPathToPlugin'`.

This can't work if my presentation is in a single HTML file. 

In my case, given I use a bundler, I'd expect to retrieve the plugin (the object with function `init`) via an ESM import and directly provide it to Reveal.js. 

I have tried to hack around the injection process with no avail.

### PDF printing

I would like to (ideally) embed both `paper.css` and `pdf.css` in my presentation so that people can print it the way they want once bundled. 

They overlap if both included in the HTML file (and wreck the presentation). 

This means that I have one of two solutions (at least that come to mind):
* either ask the parceljs team to merge a PR that would allow to retrieve the content of a CSS file in a variable (like: `import pdfCss from 'reveal.js/print/pdf.css'`) and inject it (conditionally) myself in the HTML, depending on the query parameter `?print-pdf`
* or make a change in Reveal.js that would allow both style sheets to be imported in the HTML and only applied if a specific class is added to `body` for example. Which would be done depending on the query parameter

Does that make sense? 

I'd be happy to help with any implementation if you are convinced this would bring value to the project (and I hope it does).

> Sorry for the big issue

quilicicf:
This is a PR show-casing one solution to the problems above :point_up:

quilicicf:
There is still the issue of the speaker notes to tackle but it's a good start IMO.

quilicicf:
Work has started in PRs https://github.com/hakimel/reveal.js/pull/2627 and https://github.com/hakimel/reveal.js/pull/2633. 

Stay tuned.

[...]

PR #2526, opened 2019-11-08
Bundler-friendliness

quilicicf:
This PR showcases one solution to [the issue raised here](https://github.com/hakimel/reveal.js/issues/2524).

Each commit is independent and they can (should) be reviewed separately.

## First commit

Adds a new dependency injection option that allows one to import a plugin ESM-style and inject it manually in Reveal without the need for fetching another resource.

## Second commit 

Changes style sheets in `css/print` so that it is built from a SASS file. 

This allows to scope it to `html.print-pdf` or `html:not(.print-pdf)` and always have them both in the HTML presentation.

That way, one can bundle a single file that can optionally be printed to paper or PDF.

## Third commit 

Show-cases the previous commits with a presentation that can be built to a single HTML file. 

After `npm i; npm run build`, use `npm run serve:demo` to run a local server that shows the built presentation and `npm run build:demo` to build it. 

It gives me a single-file presentation of `192K` (`56K` zipped).

## Important note

There seems to be a small bug with the PDF printing that I'm not sure I introduced, the first page's background is black (the theme is black) but the remaining pages are white. 

I'll try to understand this but if someone spots the bug...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ node_modules/
 .sass-cache
 css/reveal.min.css
 js/reveal.min.js
+/dist
+/.cache
--- a/README.md
+++ b/README.md
@@ -484,6 +484,13 @@ You can add your own extensions using the same syntax. The following properties 
 - **callback**: [optional] Function to execute when the script has loaded
 - **condition**: [optional] Function which must return true for the script to be loaded
 
+You can additionally use the following syntax, in case you are using a bundler:
+- **id**: the id of the plugin to load
+- **plugin**: the plugin object to load. It is the plugin implementation that can contain an `init` function
+- **async**: [optional] Flags if the script should load after reveal.js has started, defaults to false
+- **callback**: [optional] Function to execute when the script has loaded
+- **condition**: [optional] Function which must return true for the script to be loaded
+
 ### Ready Event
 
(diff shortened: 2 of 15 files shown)


quilicicf:
Force-pushed because the build error seems to be a weird time-out. 
Works on my machine :man_shrugging:

quilicicf:
Hey @hakimel, now that you're spending time on v4, any opinion on this topic ? 
I'd be super motivated to help you if that seems like something you want to include in the v4.

hakimel:
Hey! Makes sense to support the inclusion of plugins that are already available on the page/bundled.

I like the idea of merging the print stylesheets so we can get rid of the conditional css include in the presentation HTML. We'll need to make sure to test that it doesn't have any side-effects though. 

Can you make separate PRs for those two changes so we can review and test them separately? Thanks!

quilicicf:
Definitely! Let's roll!

quilicicf:
Here's the first one: https://github.com/hakimel/reveal.js/pull/2627

quilicicf:
And the second one: #2633 

As you said, we must make sure it doesn't break things and it does. If you spot the issue that'd help. I didn't have time to look for it, I'll try as soon as I get the time :-)

PR #2532, opened 2019-11-21
add recycle for unbinding event

Jchou24:


js/reveal.js
```diff
@@ -1267,19 +1281,7 @@
...
-				// Make sure we're dealing with JSON
-				if( typeof data === 'string' && data.charAt( 0 ) === '{' && data.charAt( data.length - 1 ) === '}' ) {
-					data = JSON.parse( data );
-
-					// Check if the requested method can be found
-					if( data.method && typeof Reveal[data.method] === 'function' ) {
-						Reveal[data.method].apply( Reveal, data.args );
-					}
-				}
-			}, false );
```
Itchibon777: window.addEventListener( 'message', function ( event ) {
				var data = event.data;

				// Make sure we're dealing with JSON
				if( typeof data === 'string' && data.charAt( 0 ) === '{' && data.charAt( data.length - 1 ) === '}' ) {
					data = JSON.parse( data );

					// Check if the requested method can be found
					if( data.method && typeof Reveal[data.method] === 'function' ) {
						Reveal[data.method].apply( Reveal, data.args );
					}
				}
			}, false );

hakimel:
Hey, we already have a `Reveal.removeEventListeners()` for this so there shouldn't be any need to add a new recycle method: https://github.com/hakimel/reveal.js/blob/master/js/reveal.js#L1559-L1595

Jchou24:
Hi, thanks for your reply

Sorry, my point is I do not find the events in recycle() to be removed in Reveal.removeEventListeners()
the events are:
	window.removeEventListener( 'load', layout, false )
	window.removeEventListener( 'load', setupPDF );
	window.removeEventListener( 'message', HandleMessage, false );
	window.removeEventListener( 'hashchange', onWindowHashChange, false );
	window.removeEventListener( 'resize', onWindowResize, false );
	window.removeEventListener( 'load', removeAddressBar, false );
	window.removeEventListener( 'orientationchange', removeAddressBar, false );

and I am not sure why the events not removed in Reveal.removeEventListeners()
so I create recycle() to avoid influencing Reveal.removeEventListeners()

hakimel:
I see. I'll review the list of events in removeEventListeners to make sure everything necessary is covered. Thanks!

PR #2534, opened 2019-11-23
Create nodejs.yml

mouse3687:


hakimel:
Thanks but we switched to GitHub actions for CI in a separate commit.

Issue #2473, opened 2019-09-02
Full setup with nodejs 12 fails


Seth-Gecko:
I'm on Debian Stretch. When installing the full setup, the build currently fails. Not sure if this is an issue with reveal.js.

[Relevant log](https://del.dog/izosojimog.rb)

For now, I have switched to nodejs 11 and everything works as expected.

xjackk:
can confirm i am getting this exact build error on node 12 as well.

edgranau:
I had a similar issue:

```
2 warnings and 5 errors generated.
make: *** [Release/obj.target/binding/src/binding.o] Error 1
gyp ERR! build error 
gyp ERR! stack Error: `make` failed with exit code: 2
gyp ERR! stack     at ChildProcess.onExit (/Users/edwin/Workspace/Edwin/Code/reveal.js/node_modules/node-gyp/lib/build.js:262:23)
gyp ERR! stack     at ChildProcess.emit (events.js:209:13)
gyp ERR! stack     at Process.ChildProcess._handle.onexit (internal/child_process.js:272:12)
gyp ERR! System Darwin 18.7.0
gyp ERR! command "/usr/local/Cellar/node/12.10.0/bin/node" "/Users/edwin/Workspace/Edwin/Code/reveal.js/node_modules/node-gyp/bin/node-gyp.js" "rebuild" "--verbose" "--libsass_ext=" "--libsass_cflags=" "--libsass_ldflags=" "--libsass_library="
gyp ERR! cwd /Users/edwin/Workspace/Edwin/Code/reveal.js/node_modules/node-sass
gyp ERR! node -v v12.10.0
gyp ERR! node-gyp -v v3.8.0
gyp ERR! not ok 
Build failed with error code: 1
npm ERR! code ELIFECYCLE
npm ERR! errno 1
npm ERR! node-sass@4.11.0 postinstall: `node scripts/build.js`
npm ERR! Exit status 1
npm ERR! 
npm ERR! Failed at the node-sass@4.11.0 postinstall script.
npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/edwin/.npm/_logs/2019-09-10T02_04_38_001Z-debug.log
```

[...]

PR #2542, opened 2019-12-03
Update node-sass to support Node.js v12

bitjson:
Install currently fails on Node.js v12, this just updates the `node-sass` dependency to support the newest Node.js versions.

(`node-sass` began support for Node.js v12 in `4.12.0` and Node.js v13 in `4.13.0`.)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -53,9 +53,9 @@
       }
     },
     "ajv": {
-      "version": "6.9.2",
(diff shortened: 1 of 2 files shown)


hakimel:
This issue was fixed in https://github.com/hakimel/reveal.js/issues/2473#issuecomment-573347475

Sorry but I didn't see this PR until after it was already fixed 🙈

Issue #1341, opened 2015-08-27
Feature request: Possibilty to mark last slide


sjoerd222888:
I have some extra backup slides in my presentation that I add in the end. I have a slide that is the last slide in the presentation (if I do not use the backup slide) and I would like to mark it as the last slide such that the progress bar at the bottom is just filled when that slide is reached. The progress bar should remain filled when going through backup slides.

I think backup slides are a common scenario where you prepare some extra slides in case there is time left or for some special question from the audience.

demoneaux:
@hakimel What do you think? IMO this use case shouldn’t be happening a lot, so I would be fine not implementing this. Additionally, a different progress bar can also be implemented simply using the `slidechanged` event.

hakimel:
@demoneaux I agree this is not something we need to add to the framework. If the progress bar not filling entirely is distracting, you can turn it off using `progress: false` in your reveal.js config.

zenzike:
I happen to also think that this would be a useful feature: backup slides are common in live presentations. The progress bar is useful to help the audience know what to expect in terms of material in the talk, so having it present is good even when there are backup slides that should not count towards progress.

zenzike:
For what it's worth, this can be achieved by making the backup slides part of the stack class. For instance:

    <section id="appendix" class="stack"> ... </section>

PR #2543, opened 2019-12-04
Allow slides to be uncounted

lassepe:
Sometimes it is useful to mark slides as uncounted, e.g. if one wishes to add backup/appendix slides but want's the progress bar to reflect the progress of "non-appendix" slides (c.f. https://github.com/hakimel/reveal.js/issues/1341). This PR adds the possibility of marking a slide with class `.uncounted` to achieve this. Currently people seem to use hacks like https://github.com/hakimel/reveal.js/issues/1341#issuecomment-237025757 which has other side affects and only works for horizontal slides.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -4206,8 +4206,11 @@
 				if( verticalSlides[j].classList.contains( 'present' ) ) {
 					break mainLoop;
 				}
-
-				pastCount++;
+        // don't count slides with the "uncounted" class attribute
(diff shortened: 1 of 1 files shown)


hakimel:
This has been merged with some changes. Instead of `<section class="uncounted">` you'll need to add an `<section data-visibility="uncounted">`. I switched to an attribute because so that we can reuse the same attribute when we eventually make it possible to fully hide a slide with `data-visibility="hidden"`.

Thanks for the PR!

Issue #2409, opened 2019-05-16
API to get width and height of Reveal object


earboxer:
There is a a `getScale()` method, there should also be

`getWidth()` and `getHeight()`,

which would be similar to `parseFloat(jQuery('.slides').css('height'))`

earboxer:
Note: `jQuery('.slides').css('height')` doesn't behave the same in all jQuery versions.

Example case: default settings, your window is any size (within the scaling limits).

```js
Reveal.getWidth() === 960
Reveal.getHeight() === 700
```

Case: width: "100%", height: "100%", your window is 1024x768
```js
Reveal.getWidth() === 1024
Reveal.getHeight() === 768
```

Case: width: 960, height: "100%", your window is 1024x768 (after #2408 is merged in)
```js
Reveal.getWidth() === 960
Reveal.getHeight() === 720
```

hakimel:
Merged in #2574

PR #2574, opened 2020-01-31
API: Expose getComputedSlideSize so plugin developers can get width and height.

earboxer:
Closes https://github.com/hakimel/reveal.js/issues/2409

Plugin developers can use `Reveal.getComputedSlideSize.width` and `Reveal.getComputedSlideSize.height` to get the width and height.

### DIFF
--- a/README.md
+++ b/README.md
@@ -638,6 +638,8 @@ Reveal.getConfig();
 // Fetch the current scale of the presentation
 Reveal.getScale();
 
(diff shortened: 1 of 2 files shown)


hakimel:
Good idea, thanks 🙌

earboxer:
Thanks for building this fantastic project, and keeping it alive!

PR #2578, opened 2020-02-05
Add RTL support to Controls arrow direction

Martinomagnifico:
Add RTL support to Controls arrow direction. Includes corrected aria label "previous" and "next".


Martinomagnifico:
Ah, this PR was against master. Will now first clone dev and re-PR...

PR #2579, opened 2020-02-05
Add RTL support to Controls

Martinomagnifico:
Add RTL support to Controls. Arrow direction and aria label will now be correct.


hakimel:
Thanks for fixing that 🎉

PR #2580, opened 2020-02-07
Allow to load highlight.js from another location

### LINKED ISSUE #2562: Upgrade highlight.js in the highlight plugin to the latest version
The highlight plugin relies on Highlight.js 9.11.0.

I think we should use the latest version available 9.17.1: https://github.com/highlightjs/highlight.js/blob/master/CHANGES.md#version-9171

Mogztter:
resolves #2562


plugin/highlight/highlight.js
```diff
@@ -74,33 +96,42 @@
 		HIGHLIGHT_LINE_RANGE_DELIMITER: '-',
 
 		init: function() {
-
 			// Read the plugin config options and provide fallbacks
 			var config = Reveal.getConfig().highlight || {};
 			config.highlightOnLoad = typeof config.highlightOnLoad === 'boolean' ? config.highlightOnLoad : true;
 			config.escapeHTML = typeof config.escapeHTML === 'boolean' ? config.escapeHTML : true;
+			config.location = typeof config.location === 'string' ? config.location : './plugin/highlight';
```
Mogztter: @hakimel is there a way to resolve a relative path from the _plugin_ directory?

test/examples/code-highlight.html
```diff
@@ -0,0 +1,66 @@
...
+
+			Reveal.initialize({
+				history: true,
+				transition: 'linear',
+
+				highlight: {
+					// highlightOnLoad: true,
+					// escapeHTML: true,
+					// location: 'https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@9.18.0/build',
+					location: '../../plugin/highlight'
```
Mogztter: Ideally this should not be needed as the default location should be relative to the _plugin_ directory.

PR #2609, opened 2020-02-27
Don't include test directory in npm package

stefnotch:
This ignores the `test` directory in the npm package and reduces the package size.


hakimel:
Makes sense, thanks!

PR #2627, opened 2020-03-09
Add bundler-friendly dependency injection

quilicicf:
## What this PR does

It adds a new dependency injection option that allows one to import a plugin ESM-style and inject it manually in Reveal without the need for fetching another resource.

## Example of it in action

```js
import Reveal from './js/reveal';
import zoomPlugin from './plugin/zoom-js/zoom-for-bundler';
import helloWorldPluginCreator from './plugin/helloWorldPluginCreator';

window.Reveal = Reveal;

Reveal.initialize({
  hash: true,
  dependencies: [
    { id: 'zoom', plugin: zoomPlugin, async: true },
    { id: 'hellows', plugin: helloWorldPluginCreator(83, 'S', 'sync plugin'), async: false }, // Loaded from an ESM plugin
    { id: 'hellowa', plugin: helloWorldPluginCreator(65, 'A', 'async plugin'), async: true }, // Can be loaded twice if that makes sense
  ]
});
```


quilicicf:
Damn, a conflict. 
I'll fix that during lunch.

hakimel:
Thanks for this! I'm making some pretty significant changes in the dev branch with the switch to modules so it's hard to avoid conflicts. No worries if you don't have time to fix them—I can handle it when merging.

hakimel:
Merged this 🙌

I wonder if we could remove the "id" from the dependency object. Existing plugins already register themselves, and their ID via, `Reveal.registerPlugin( 'zoom', RevealZoom )`. What do you think @quilicicf?

quilicicf:
Well, I think the fewer fields we have the better but how would you define the plugin's id when registering it via `Reveal.initialize({ dependencies: [ { thePlugin } ] })` then? :thinking:

hakimel:
Never mind. I dug into the code and what I was picturing would not work so let's leave the ID in. 

One alternative approach to including bundled plugins that should work is:
```javascript
Reveal.registerPlugin( 'pluginID', pluginInstance )
Reveal.initialize(...)
```

quilicicf:
Indeed, but the registration via `initialize` being documented, I supposed we're forced to support it :-(

Issue #2537, opened 2019-11-26
Bug: Shift-Space does not navigate back when speaker notes layout was changed


thomham:
Keyboard shortcut `Shift-Space` does not navigate back when speaker notes layout was changed at least once. Instead it navigates forward as if the Shift-key were not pressed.

Reproduced on macOS and Windows in Google Chrome in revealjs version 3.8.0.

Workaround is to use keyboard shortcut `p`.

Issue #1249, opened 2015-06-02
Shift + Space does not work in speaker controls


estan:
Pressing `Shift + Space` to go back works in the presentation itself, but not from the speaker remote controls. The arrow keys work of course, but then I have to remember how my sections are nested. With `Shift + Space` I can get back, whether that is up or left, so it would be great if this worked also from the speakers remote controls.

estan:
To be clear what I mean: `Shift + Space` in the presentation goes to the previous slide (works the same as `p`). In the speaker remote controls, it goes forward (same as just `Space`). I'd expect `Shift + Space` to work the same in both (work as `p`).

estan:
Bah. And all of a sudden it seems to work. I'm sure it didn't just a while ago, so I'll keep this open for a little while in case I can reproduce it again.

estan:
Okay. Right now it's not working again. `Shift + Space` goes forward. Not sure what triggered it though :/

estan:
My previous analysis was wrong. I can get this behavior even if i switch focus to the presentation and back again. So still not sure what triggers the bug :/

lagartoflojo:
To make `Shift + Space` work when the speaker notes are focused, you need to click on the current slide preview.
Clicking on the time or speaker notes breaks `Shift + Space`.
Clicking on the Upcoming preview will cause both `Space` and `Shift + Space` to only go to the next/previous slide in the Upcoming pane, not the current pane nor the actual presentation.

jmuheim:
I still have problems with this. I'm tempted to only use `p` and `n`, as they always seem to work properly... but it would be nice to be able to use `space`, still. Any progress on this?

jmuheim:
I stumbled over exact this problem again, one year later.

And I don't seem to be the only one:

- https://github.com/hakimel/reveal.js/issues/2537
- There's even a PR for fixing it: https://github.com/hakimel/reveal.js/pull/2629

Any chance that we get this annoying bug fixed? I asked Claude.ai to fix it in my local copy and let it create a summary of it (but it would obviously be way better to fix this problem once and for all):

## Bug

In the main presentation window, `Space` advances to the next slide and `Shift`+`Space` goes back to the previous slide. In the **speaker notes** popup (opened with `s`), `Shift`+`Space` incorrectly also advances to the **next** slide instead of going back.

## Root cause

[...]

PR #2629, opened 2020-03-09
Fixing shift key not working in slide notes bug

telliott22:
Fixing issue #2537. Passing shiftKey event property back from notes window with keyCode


jmuheim:
I hope this can be fixed soon, Claude.ai took a few moments to find the problem and solution, see https://github.com/hakimel/reveal.js/issues/1249#issuecomment-4772870823

PR #2633, opened 2020-03-10
Bundler friendliness (part II) - Repackage print style sheets

quilicicf:
## What this PR does

It applies paper.css and pdf.css conditionally (depending on whether the query parameter `?print-pdf` is defined) so they can both be loaded in the presentation without adding a script.

The style sheets can now be both imported and only apply on the following conditions:

* `paper.css`: only applied when printing (media query) and the query parameter is NOT present
* `pdf.css`: only applied when the query parameter is present

## Question

Shouldn't the pdf style sheet also be applied only when printing? I believe so but I'm not certain.

## :warning: The current version is a bit buggy

I didn't validate using the test suite because it is commented out (see gulpfile). 

I did inspect the result on the page demo.html though and found a nasty bug that I don't have the time to locate right now. 

The first slide when printing to pdf with black theme follows the theme correctly but all the other slides have a white background. 

I'll try to fix this ASAP, don't merge in the meantime :-)


quilicicf:
Soooo, I have checked the CSS bug for PDF printing and it appears to be in the `dev` branch too. 
For an unknown reason, the option `background graphics` in Chrome print preview was auto-activated when testing on your branch but not on mine :thinking: 

Turns out the backgrounds are OK on both branches when turning it on. 

There's one remaining issue (on both branches too), a page is added when printing to PDF, and given it has no content, it also has no background => it is white even with dark themes.

quilicicf:
Dug a bit further. The additional page in the PDF is actually added because there's a tiny DOM element that gets added at the end of the presentation that looks like this:

```html
<div class="aria-status" aria-live="polite" aria-atomic="true" style="position: absolute; height: 1px; width: 1px; overflow: hidden; clip: rect(1px, 1px, 1px, 1px);">
    Presentation's first slide's text here
</div>
```

Switching the height to 0 removes the additional page.

quilicicf:
I have added a commit to hide the status element off-screen. 
It works for me now. 
Can you check that I didn't wreck anything @hakimel ?

js/reveal.js
```diff
@@ -368,7 +370,7 @@ export default function( revealElement, options ) {
 		// Limit the size of certain elements to the dimensions of the slide
 		createStyleSheet( '.reveal section>img, .reveal section>video, .reveal section>iframe{max-width: '+ slideWidth +'px; max-height:'+ slideHeight +'px}' );
 
-		document.body.classList.add( 'print-pdf' );
+		document.querySelector('html').classList.add( 'print-pdf' );
```
hakimel: This should use `document.documentElement` instead of `querySelector('html')`, to match how the rest of reveal.js accesses it.

js/reveal.js
```diff
@@ -292,6 +292,8 @@ export default function( revealElement, options ) {
 			statusElement.style.position = 'absolute';
 			statusElement.style.height = '1px';
 			statusElement.style.width = '1px';
+			statusElement.style.top = '-1px';
+			statusElement.style.left = '-1px';
```
hakimel: The status element is used for accessibility—it calls out text to a screen reader. It serves no purpose in PDF exports though, so maybe we can add `.aria-status { display: none; }` to the pdf print sheet instead of changing it here?

quilicicf: Yup, seems reasonable. I'll do that.

js/reveal.js
quilicicf: Will do

quilicicf:
Comments taken into account.
Another thing I saw is that the comments stay in the CSS after SASS compilation. And they jump to the top of the file, meaning they have no context. 

Shouldn't they disappear in the CSS? I can configure the gulp task but I want to be sure what's expected here. 

Example, pdf.css starts like this:

```css
/* Copyright... */

html.print-pdf {
  /* Remove any elements not needed in print. */
  /* Slide backgrounds are placed inside of their slide when exporting to PDF */
  /* Display slide speaker notes when 'showNotes' is enabled */
  /* Layout option which makes notes appear on a separate page */
  /* Display slide numbers when 'slideNumber' is enabled */
  /* This accessibility tool is not useful in PDF and breaks it visually */ }
```

I don't think Reveal should add the comments in "production" files, should it?

hakimel:
PR is looking good so I'm gonna start testing.

I'll take a look at the build to see why it's not removing those comments. The whole dev branch—including the new gulp build—is undergoing some pretty big changes so it's not in a stable place yet :)

hakimel:
Also one thought that springs to mind looking at this is why don't we just bundle the pdf and paper styles inside of reveal.css to avoid the separate imports? I'll poke around and see if that's feasible while testing.

quilicicf:
I can definitely try to merge them if that helps, just tell me. 
I was also wondering if the pdf thing should not be also in the print media query ? 
I'm not sure it serves a purpose to wreck the style of the presentation when the `?print-pdf` parameter is there and the user is not trying to print.

hakimel:
Let's avoid wrapping the PDF styles in a media query. There are multiple scripts out there for exporting reveal.js presentations to PDF, including our implementation at Slides.com, that may be affected by that.

I'm experimenting with building the print styles into the main stylesheet—will let you know how it goes.

quilicicf:
OK, good luck :)

hakimel:
Merged with two changes;
1. Paper and PDF print styles are now baked into reveal.css 🥳 Updated README accordingly.
2. The print to paper styles were severely outdated so I updated them.

quilicicf:
Noice. Will attack the third part of my evil plan next week. 
I need to make the speaker notes plugin bundler-friendly which won't be a piece of cake. 
I'll share my ideas first as this is gonna be more complex by far than the two first parts IMO.

PR #2636, opened 2020-03-20
Update README.md

julianofischer:
Added blood and moon themes to the theme list.


hakimel:
Thanks, I've added those two to the docs. (The docs moved to https://github.com/reveal/revealjs.com as of 4.0)

PR #2647, opened 2020-04-17
Add gitpod config

Truta446:
this commit adds support for Gitpod.io, a free automated
dev environment that makes contributing and generally working on GitHub
projects much easier. It allows anyone to start a ready-to-code dev
environment for any branch, issue and pull request with a single click.


hakimel:
Thanks for the PR but no thanks. There are many good options for people who want to clone reveal.js into a cloud editor, I don't think we need to pick one or the other.

PR #2651, opened 2020-04-23
reveal.js 4.0.0

hakimel:
I've been hard at work on reveal.js 4.0.0 for a while now and wanted to share the progress so far. It's not ready to merge just yet, but it's getting close.

The main changes in this release are:
- New website and docs (WIP) https://revealjs.netlify.com
- Modernized reveal.js core. The old reveal.js file with 6.2k lines of code (!) was refactored and split into ES modules.
- You can now run multiple reveal.js presentations on the same page.  
  ```javascript
  import Reveal from 'reveal.js'
  let deck1 = new Reveal( document.querySelector( '.deck1' ) );
  let deck2 = new Reveal( document.querySelector( '.deck2' ) );
  deck1.initialize({ embedded: true });
  deck2.initialize({ embedded: true });
  ```
- Converted all plugins to ES modules so that they can be bundled with reveal.js.
- All plugins and reveal.js core are transpiled to ES5 for situations where you need to support older browsers.
- Switched build systems from grunt to gulp (using rollup for bundling).
- Added a major new Auto-Animate feature.
- Added a new `slidetransitionend` event.
- `Reveal.add/removeEventListener` -> `Reveal.on/off`.
- Added support for code line numbers and highlights using Markdown
- The initialize method now returns a promise that resolves when reveal.js has started ```Reveal.initialize().then( () => console.log( 'ready' ) )```.
- The highlight.js and marked libraries are now installed from npm. They were previously hard coded inside their corresponding plugins in this repository.

There are a few breaking changes in this release:
- All compiled assets (reveal.js, reveal.css, themes) have moved to /dist.
- Print stylesheets are now baked into the reveal.css file so there is no need to separately include them in the presentation HTML.
- The multiplex and server side speakers notes plugins have moved out to separate repos and can be installed via npm:
  https://github.com/reveal/multiplex 
  https://github.com/reveal/notes-server

Docs, upgrade instructions and a new revealjs.com will follow when it's ready to merge but hopefully this gives you an idea of what's to come! ✌️


README.md
```diff
@@ -18,14 +18,15 @@ This project was started and is maintained by [@hakimel](https://github.com/haki
 - [Initialization](#initialization)
 - [Configuration](#configuration)
 - [Presentation Size](#presentation-size)
-- [Dependencies](#dependencies)
 - [Ready Event](#ready-event)
```
avelasquez29, line 21:
```suggestion
- [Ready Event](#ready-event)
```

avelasquez29, line 21:
```suggestion
- [Ready Event](#ready-event)
```

rjarry:
Hi @hakimel , pdf print seems broken on the `dev` branch.

All I have is a trace from chromium console:

```
print.js:44 Uncaught TypeError: this.Reveal.layoutSlideContents is not a function
    at value (print.js:44)
value @ print.js:44
load (async)
ve @ reveal.js:214
Promise.then (async)
ue @ reveal.js:135
Y.initialize @ index.js:38
(anonymous) @ ?print-pdf:31
```

hakimel:
@rjarry Thanks for letting me know—I tried it out and saw a slightly different exception from your trace and have fixed that. Can you give it another shot when you get a chance and let me know if it works?

If it's still broken and you're able to share the presentation HTML that'd really help.

rjarry:
@hakimel thanks it works now!

McShelby:
Will switching of the stylesheets for PDF export now handled internally? Just asking to be prepared with my plugins.

Additionally, could you give an id attribute to the link element of the syntax highlight stylesheet? This would make things easier - at least for me but also for other plugin writers.

hakimel:
@McShelby Yes, the PDF print styles are now baked into the dist/reveal.css file. No more need for this hack in the HTML 🥳 https://github.com/hakimel/reveal.js/blob/00b0ace3863952ce95df4430d203686fd27bfc20/index.html#L16-L23

Sure thing—I just pushed the ID "highlight-theme".

I've made a few changes to improve the plugin workflow in 4.0. The docs are still in progress, but you can see the current version here: https://revealjs.netlify.app/plugins/

McShelby:
Thanks, you're truly awesome!

McShelby:
Some observations while starting to port my plugins to reveal 4:

- Running the presentation from a file:// URL leads to CORS blocking.
- Switching to the old module syntax in demo.html is possible (and solves CORS) - but in that case the exported globals are usually prepended with "Reveal" eg. "RevealZoom". This makes it necessary to further tweak demo.html which is a bit cumbersome. Better let both variants use the same name.
- Having all plugins flat in the dist/plugin directory may get a bit messy - at least if a plugin needs further files besides its *.js file.

McShelby:
...and
- `addKeyBinding` does not add to the help screen (eg. speaker notes shortcut not displayed)

hakimel:
Thanks for the feedback! I've fixed and added tests for the key binding error. I'm considering switching away from the ES module reveal.js version in demo.html so people aren't confused about the origin errors when opening from file://.

The ES module plugins use default exports so you can call them anything you'd like.
```js
import rvl from '/dist/reveal.esm.js';
import md from '/dist/plugin/markdown.esm.js';
rvl.initialize({ plugins: [ md ] });
```

The ES5/umd versions need to have non-conflicting names since they are available in global scope, hence `RevealMarkdown`, `RevealHighlight` etc.

No more plugins will be added to this repo so I'm not concerned about /dist/plugin growing. I've moved two plugins out to their own repos in 4.0 and will probably separate more of them down the line. That said, maybe it makes for a nicer URL to include from `/plugin/highlight/highlight.esm.js` instead of `/dist/plugin/highlight.esm.js`?

Btw, I just published the first version of the docs for creating plugins: https://revealjs.netlify.app/plugins/creating-plugins/

McShelby:
>  I'm considering switching away from the ES module reveal.js version in demo.html so people aren't confused about the origin errors when opening from file://.

I would appreciate this, cause I hate if webpages need a server to work correctly. Especially if you "only" want to display an self-contained offline presentation.

On the other hand it's sad to not be able to show off ES modules in the demo. So maybe two demos? Just thinking out loud :-)

> The ES5/umd versions need to have non-conflicting names since they are available in global scope, 

But only if the page is loading the same plugin as ES module AND umd. This should not be the case in a properly set up presentation - or do I miss a thing here?

> No more plugins will be added to this repo so I'm not concerned about /dist/plugin growing.

I was more thinking of the case, that user installs addional plugins into the presentation. But maybe that's not the directory they are supposed to be. Is there a recommendation I overread?

Btw: I'm done porting my plugins to reveal 4. Each plugin is capable to run with both reveal vesions (if they are loaded by using a script tag).

hakimel:
Glad to hear you were able to make your plugins work with both versions!

There's currently no recommendation for where third party plugins and assets should be placed. Some plugins may be installed via npm and others copy-pasted from the file system. If a presentation is being copied into the repo I think /plugin/<pluginName>/ is a good place for it.

hakimel:
reveal.js is finally getting a logo. I'm still working on minor refinements, but this is the general direction!
Screen Shot 2020-05-13 at 15 04 20

Martinomagnifico:
Hi Hakim, cool progress you're making! Love the auto-animate :-) Also, the slidetransitionend is something that my Appearance plugin can use. 

Could you also emit an event for fragmenttransitionend? For example, if a user has set a slow css transition for a fragment, and there is a chart in there, then it can hold off to start building the chart columns when the fragment transition is complete. 

Having a built-in fragmenttransitionend would then make my Transit plugin, currently needed for Appearance, completely obsolete, which is something we talked about earlier. I would prefer that to asking people to keep using it for fragment transitions.

hakimel:
Thanks @Martinomagnifico! An event for fragment transitions could be useful too. It's not something I'm going to add to 4.0 since I'm trying to wrap that up for release but will consider it for another update.

hakimel:
I've moved the built plugins from `/dist/plugin/math.js` to `/plugin/math/math.js`. Been going back and forth on this for ages, but I think it's better to isolate everything plugin related to their original folder. Also, as @McShelby pointed out, it makes it more obvious where to place additional plugins (/plugin) when copy-pasting them into a reveal.js directory.

Also the notes plugin is now 100% bundled in a single notes.js file. It used to require loading of an external notes.html HTML file, which led to issues with changing relative paths etc.

hakimel:
4.0.0 is now merged into master 🚀 If anyone wants to try it out and give feedback I'd really appreciate it: https://github.com/hakimel/reveal.js/releases/tag/4.0.0-rc.1

Release notes are underway and upgrade instructions can be found here https://revealjs.netlify.app/upgrading/

McShelby:
So far everything works fine for me.

Just another small wish: At the end of index.js you are providing a pre 4.x compatiblity layer. Could you add the VERSION attribute so a plugin is not required to do duck typing if it needs special logic for certain versions during its loading time?

RealCyGuy:
@hakimel this upgrading link works: https://revealjs.com/upgrading/

hakimel:
Thanks for testing!

@McShelby Great suggestion, `Reveal.VERSION` is now immediately available.

@RealCyGuy I switched revealjs.com over the new website today. DNS might still be propagating depending on where you're located though.

If anyone wants to contribute to the docs, the repo for is https://github.com/reveal/revealjs.com.

Martinomagnifico:
Hi Hakim, there is something I noticed regarding Themes:

The white and black theme react to section(data-background="some other color") when this color is dark or light compared to the main slide background. It's a CSS rule that in those cases sets the font color to a contrasting color. The other Reveal themes do not have such a rule. So if a user sets a very contrasting background in any of the darker themes (compared to the main background), the fonts become almost unreadable. The good thing is that the controls do change to a contrasting color. Also, the colors set in the rules in Black and White directly use a Hex color, not a variable. I would suggest adding exposed variable colors like --main-color-opposite, --heading-color-opposite etc., or at least add rules in those themes for dark/light backgrounds.

hakimel:
@Martinomagnifico Thanks for the feedback! I've made a note to add inverse text colors to the rest of the themes as well and expose them as CSS variables. This isn't something that has changed in version 4.0 so we'll save it for next version.

hakimel:
reveal.js 4.0.0 is now shipped and available from npm 🚀

Release notes: https://github.com/hakimel/reveal.js/releases/tag/4.0.0

zxq08:
想请问一下，关于4.0的npm版本，**支持的最低node版本**有介绍吗，抱歉实在没有找到...

RatStar811 (approved):

PR #2666, opened 2020-05-19
fixed links

RealCyGuy:
is this intentional? because the links don't work


hakimel:
Oops, no. I was replacing revealjs.netlify.app with revealjs.com before and screwed it up. Thanks for the fix!

PR #2671, opened 2020-05-21
Fix slide notes printing

s-l-lee:
The change to `print.js` fixes an `Uncaught ReferenceError: getSlideNotes is not defined` when attempting to export to PDF on Chrome with `?print-pdf` in the query string and `showNotes` set to `true`, `'inline'`, or `'separate-page'` in the configuration.

The change to `notes.js` fixes the notes sidebar remaining visible when attempting to export to PDF on Chrome under similar conditions as shown in the screenshot below:
print-preview


hakimel:
Thanks for spotting and fixing that 🙌

Issue #2559, opened 2020-01-04
MathJax 3.x compatibility issue


wirepatch:
MathJax 2.x uses URL configuration parameters e.g.:
```
https://cdnjs.cloudflare.com/ajax/libs/mathjax/2.7.5/MathJax.js?config=TeX-AMS-MML_CHTML
```
In current 3.x this dispatch mechanism has been superseeded using e.g.:
```
https://cdn.jsdelivr.net/npm/mathjax@3/es5/tex-svg.js
```
This works using the reveal MathJax plugin e.g.:
```
math: {
    mathjax: 'mathjax/tex-mml-svg.js',
    config: 'x' // Obsolete dummy value.
}
```
However using »Inspect« Chromium shows the following JavaScript error:
```
math.js:71 Uncaught TypeError: Cannot read property 'Config' of undefined
    at math.js:71
    at HTMLScriptElement.finish (math.js:43)
```
Omitting `config: 'x' ` completely does not help either.

Technically it's a mere nuisance. But if you are chasing for errors this one is confusing. Making »config« an option rather than a mandatory property would help.

burgerga:
MathJax 3 support would be nice, but it would require almost a complete rewrite of https://github.com/hakimel/reveal.js/blob/master/plugin/math/math.js

I use markdown slides and trying to use Mathjax 3 (also with the `config: 'x'` trick) breaks most (but not all) of my equations.
The reason is that Mathjax v3 is loaded and configured differently than v2 (compare http://docs.mathjax.org/en/v2.7-latest/configuration.html and http://docs.mathjax.org/en/latest/web/start.html), that is why you get this 
```
math.js:71 Uncaught TypeError: Cannot read property 'Config' of undefined
    at math.js:71
    at HTMLScriptElement.finish (math.js:43)
```
error, which is not a "mere nuisance"; it means that most of the [math plugin](https://github.com/hakimel/reveal.js/blob/master/plugin/math/math.js) is not working (i.e. none of your options are applied and the reveal event listener is not added). Additionally, v3 doesn't use queues so that part needs to be rewritten as well.

burgerga:
Instead of overwriting the current plugin, I made a new one, because at the moment there are still some issues with MathJax 3 (I encountered [newlines not working](https://github.com/mathjax/MathJax/issues/2312) and [under/over braces not working](https://github.com/mathjax/MathJax/issues/2402)), so if you want to change back to MathJax 2 just enable the other plugin again.

[...]

PR #2677, opened 2020-05-25
Replace Mathjax 2 with Katex

burgerga:
As discussed in #2559 

First I dynamically inserted the katex javascript sources using defer, but that only worked half of the time, because they were loaded in random order (I still don't get why). So instead I chained them so that the auto render js gets loaded on the `onload` of the main katex script. 

It's possible to specify versions of katex, eg (default):
```js
math: {
  version: '0.11.1'
}
```

but also 

```js
math: {
  version: 'latest'
}
```

In addition you can use a local katex, for example after `npm install katex` you can use
```js
math: {
  local: 'node_modules/katex'
}
```


burgerga:
TBH it might just be easier (for us :wink:) if people used the instructions from https://katex.org/docs/autorender.html#usage

hakimel:
This is looking great, thanks! Will merge once I’ve had a chance to try it out more thoroughly.

burgerga:
I see I might have forgotten a call to `deck.layout` after `renderMathInElement`, will fix that in a moment

burgerga:
Now with more efficient `async await` script loading that allows mhchem to be inserted easily in the middle.
I can now also do the `renderMathInElement` call directly in the `.then` which, for some reason, I couldn't do before with the chained `loadScript` calls.

burgerga:
Maybe we can load mhchem support by default, it's hardly noticeable performance-wise, and it's pretty cool.

rodrigoalcarazdelaosa:
This didn't make it to version 4 right?

hakimel:
@rodrigoalcarazdelaosa nope, though the plan is to include it in 4.1.

I received some feedback about possible negative side-effects of switching away from MathJax so I'm still thinking about how to best make this transition. It's hard to tell exactly what the tradeoffs are when I don't use any math typesetting myself. I'll dig back into this soon once I resume work on 4.1 after vacation.

rodrigoalcarazdelaosa:
The [Academic theme](https://github.com/gcushen/hugo-academic), which I use, transitioned a while ago from v2 to v3. Maybe [this thread](https://github.com/gcushen/hugo-academic/issues/1400) can help.

brichwin:
To make math content accessible, MathJax is the way to go. It's support for non-visual access to mathematical content is very mature and robust. Also, MathJax v3 now supports braille math output which is SUPER awesome. Imagine being forced to have to listen to long complicated formulas. If you move to MathJax v3, then users with braille displays can use them to "sight read" by touch the online math just like the rest of us.

MathJax also has features that support those with mild vision loss and individuals with print impairments. 

Many of us working in environments where accessible content is the law depend on MathJax to get our STEM / Mathematics content accessible online. 

From what I can see on issues open the Katex github their support for assistive technology is far from mature or even being considered working.

The American Mathematical Society (AMS) backs the MathJax tool and the accessibility extensions are now included by default in MathJax v3. 

Please seriously consider switching to MathJax v3 instead of Katex.

otaconjh:
Katex supports MathML and is committed to continue supporting it [for accessibility reasons](https://github.com/KaTeX/KaTeX/issues/593#issuecomment-323553223)
To my understanding, Katex and Mathjax both support MathML

brichwin:
Katex has many bugs in their rendering that they have yet to resolve after years of knowing of the issues. See: https://github.com/KaTeX/KaTeX/issues/820 for one and https://github.com/KaTeX/KaTeX/issues/38 for another.

MathJax has no such issues. I've encountered another project where Katex was used recently and our screen-reader users could not access the math rendered by Katex.

It seems real world experience doesn't favor Katex.

burgerga:
@brichwin I am a very happy KaTeX user, and the main benefit for me is the fast rendering times. But opinions (and needs) differ, so [what I proposed originally](https://github.com/hakimel/reveal.js/issues/2559#issuecomment-630370634) is a separate `reveal.js-math` repo, where we support MathJax v2, v3 and KaTeX. I am still open to this, but it's ultimately @hakimel 's decision since he wrote the Mathjax v2 plugin, and my Mathjax v3 is based on that.
If you want to use Mathjax V3, I wrote a plugin for reveal.js <4.0 (https://github.com/hakimel/reveal.js/issues/2559#issuecomment-629609465), which is very easy to adapt to the new plugin format required by reveal.js >= 4.0

PR #2682, opened 2020-06-02
allow for custom targets for the slide numbers

SirkoS:
Currently you can just switch slide numbers on or off. I think it would be a rather nice feature to allow users to place the slide numbers into custom locations.

In particular, I have a header for my slide deck, that should include the current slide number. With the included changes, I can add a place holder (as follows) at the location I want the slide numbers to show up.

```html
<span class="slide-number-target"></span>
```

### DIFF
--- a/js/controllers/slidenumber.js
+++ b/js/controllers/slidenumber.js
@@ -11,9 +11,17 @@ export default class SlideNumber {
 
 	render() {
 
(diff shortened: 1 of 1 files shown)


SirkoS:
solved by plugin

vivtory37:
Thank you, i appreciate everything you do.

PR #2746, opened 2020-08-04
Fix URL to pdf-export documentation

ebriand:
Hey 👋 

The URL to the pdf-export documentation in the pdf.scss file was incorrect, this small PR fixes it 😄


hakimel:
🙌

PR #2752, opened 2020-08-18
Start relative paths in CSS with ./

michaelmior:
The bundler I'm using happens to not work well with relative paths that don't start with `./`.
Since this will not break anything existing, I've added this prefix in the themes and font CSS files.


hakimel:
Thanks!

michaelmior:
@hakimel Thanks for the merge! Looking forward to the next release :)

PR #1915, opened 2017-06-04
Fix speaker note separator regex

jackycute:
it should only take effect on the line start
related: https://github.com/hackmdio/hackmd/issues/360

PR #2762, opened 2020-08-28
More reasonable notes separator

KmolYuan:
I think the default notes separator should be starts in first column to avoid inline usages.
That's inconvenient to use an escape character or override it with a custom option.

```markdown
... my notes: "Remember to ...
... my notes&colon; "Remember to ...
```


mattcen:
Duplicate of #1915 (albeit using newer base regex that has an optional 's' at the end)

PR #2771, opened 2020-09-07
docs: Fix simple typo, transiition -> transition

timgates42:
There is a small typo in js/controllers/autoanimate.js.

Should read `transition` rather than `transiition`.


bryanleejohnson:
thanks for fixing the problem we are part of a great team. keep up the good
work.

PacMani3 (approved):

PR #2799, opened 2020-10-05
Update to demo.html

theumang100:
meta tag added


the-solipsist:
This is #Hacktoberfest spam.  

The same meta tag already exists on line 15 of the file:
https://github.com/hakimel/reveal.js/blob/0582f57517c97a4c7bfeb58762138c78883f94c5/demo.html#L15

PR #2858, opened 2020-11-30
Fixed path to themes in demo.html

famura:
The paths to the different themes are not found, causing the demo presentation to fall back to the default theme when clicking on the links in the section `themes`.


hakimel:
Good catch, thanks!

PR #2890, opened 2021-02-03
Implement data-markdown-text attribute

cpmech:
Sometimes it's very useful to set the markdown text directly, instead of loading a file (e.g. via ajax).

This situation happens, for example, when using React components.

This pull-request implements a **data-markdown-text** attribute that receives an **escaped** string representing the markdown text.

### DIFF
--- a/examples/markdown.html
+++ b/examples/markdown.html
@@ -18,6 +18,9 @@
 
 			<div class="slides">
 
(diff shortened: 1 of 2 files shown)


cmoreno323 (commented):
Merge

yubele:
👍

PR #2901, opened 2021-02-13
Expose preview management functions

sestegra:
After upgrading to 4.x, preview management functions are missing from API.

I'm using them to show iframe in preview mode.


hakimel:
Makes sense, thanks!

PR #2933, opened 2021-04-11
Use computed background style if colorToRgb fails (fixes #2635)

### LINKED ISSUE #2635: has-dack-background not applied when background-color is a color name
Although valid HTML, using `background-color=black` on a slide will not trigger the `has-dark-background` to be applied by reveal.js. However, using `background-color=#000` will work.

The problem seems to be that `colorToRgb` doesn't support named colors. All strings sent in that function get analyzed as hex-encoded color values.

Work-around: refrain from using named colors and use their hex-encoded version instead.

csachs:
The `has-dark-background` class is not properly added, if a named color (e.g. `black`) is used for the `data-background-color` attribute.

If the attribute is explicitly set, it will be passed to `colorToRgb()` which can only handle explicit rgb notations, not named colors.

As a simple workaround, this PR tries the conversion beforehand, and will use the computed styles if it fails.

Alternatively, one could expand `colorToRgb()` to support named colors, e.g. by querying the browser via a `canvas` (c.f. https://stackoverflow.com/a/47355187 ).

### DIFF
--- a/js/controllers/backgrounds.js
+++ b/js/controllers/backgrounds.js
@@ -192,8 +192,8 @@ export default class Backgrounds {
 		// color, no class will be added
 		let contrastColor = data.backgroundColor;
 
-		// If no bg color was found, check the computed background
-		if( !contrastColor ) {
+		// If no bg color was found, or it cannot be converted by colorToRgb, check the computed background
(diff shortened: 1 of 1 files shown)


hakimel:
Good catch. Thanks for the fix!

PR #2940, opened 2021-04-23
Fix: multiple image-urls in data-background-image

cmeessen:
Currently, when using the `data-background-image` attribute, the user can only specify one background image. This is, because if a url is given, the whole string given in `data-background-image` is wrapped into `url()`.

This MR allows to use a comma-separated list, e.g.:

```html
<section data-background-image="image1.png, image2.png"></section>
```

Would be happy to see this merged into `dev`.

### DIFF
--- a/js/controllers/slidecontent.js
+++ b/js/controllers/slidecontent.js
@@ -102,7 +102,13 @@ export default class SlideContent {
 
 				// Images
 				if( backgroundImage ) {
-					backgroundContent.style.backgroundImage = 'url('+ encodeURI( backgroundImage ) +')';
+					let backgroundString = ''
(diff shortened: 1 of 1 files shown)


cmeessen:
This bug also exists in v3.9. If this version is still maintained, I would also create a MR for this branch.

hakimel:
Looks good, thanks! There's no reason to patch the 3.9 branch.

ArtNow198:
> Looks good, thanks! There's no reason to patch the 3.9 branches.

sultan801 (changes requested):
Hello

faxm0dem:
When I use this feature, only the first image is being displayed. Is there supposed to be a slideshow of backgrounds ?

PR #2989, opened 2021-07-28
Fix imgs

brownierin:
Messed with the CSS, turns out my linkrel was the wrong path. Whoops!


brownierin:
Sorry! This was supposed to go to my fork.

PR #3001, opened 2021-08-30
Fix unmatched h3-tags in example

poacher2k:
Also present in [the test](https://github.com/hakimel/reveal.js/blob/01d8d669bc2b681b595262ccbe27293eec2fcb44/test/test-auto-animate.html#L26-L32)


Pakon2543:
``def fibonacci(n)
  return n if ( 0..1 ).include? n
  (fibonacci(n - 1) + fibonacci(n - 2)) #recursive calls
end

Pakon2543:
def fibonacci(n)
  return n if ( 0..1 ).include? n
  (fibonacci(n - 1) + fibonacci(n - 2)) #recursive calls
end

hakimel:
Thanks, your commit has been merged bc5ca9a452fd2b3e72b069a0a0021639cd3eebad

PR #3005, opened 2021-09-09
Update comments about markdown slide separators

dabrahams:
They might have been correct once upon a time, but apparently not since "---" became the default.


hakimel:
Thanks!

PR #3010, opened 2021-09-16
Support multiple aside notes elements per slide

lechten:
So far, multiple notes per slide are only supported if they are
attached to fragments (without fragments, only the first aside notes
element on a slide is displayed).  With this commit, the contents of
all notes on a slide are displayed (except for fragments with notes,
for which, as before, only each fragment's first/single note is
displayed).


ramonastaudhammer1 (approved):

lechten:
I'm wondering: Should I also include the generated/minified JavaScript files in this PR? That might make it easier for others to test my fork, but I'm not sure about conventions for this case...

lechten:
I went ahead to add generated files to simplify testing for others.

hakimel:
This has been merged https://github.com/hakimel/reveal.js/commit/f9ce61e34af22458c7a38bda4084513c566fcee4

Thanks!

Issue #2761, opened 2020-08-28
Register new language definition in highlight plugin


rajgoel:
I want to add a language for code highlighting that is not available in the highlight plugin. Without the plugin I manage to make it work as follows:

```html
<!doctype html>
<html lang="en">

<head>
	<meta charset="utf-8">
	<meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no, minimal-ui">

	<link rel="stylesheet" href="../reveal.js/dist/reveal.css">
	<link rel="stylesheet" href="../reveal.js/dist/theme/black.css" id="theme">
	<link rel="stylesheet" href="../reveal.js/plugin/highlight/zenburn.css">

	<script src="reveal.js/dist/reveal.js"></script>
<!--	<script src="reveal.js/plugin/highlight/highlight.js"></script> -->
<script src="https://cdn.jsdelivr.net/gh/highlightjs/cdn-release@9.12.0/build/highlight.min.js"></script>

</head>

<body>

<div class="reveal">
<div class="slides">
<section>
<pre><code class="ampl">
var h;
/*
A comment
*/
var h;

# Part 1:  Variable Declaration (var, set, param, etc)
maximize var BluePaint;
var BlackPaint;
# Part 2:  Objective Function

maximize Revenue:  10*BluePaint + 15*BlackPaint;
# Part 3:  Constraints 
subject to Time:   (1/40)*BluePaint  +  (1/30)*BlackPaint<= 40;
subject to BlueLimit:  0<= BluePaint<= 1000;
subject to BlackLimit:  0<= BlackPaint<= 860;

</code></pre> 
</section>
</div>
</div>


<script>
Reveal.initialize({
//	plugins: [ RevealHighlight ],
});

Reveal.addEventListener('ready', (event) => {
	hljs.registerLanguage("ampl",function(e){var r="([a-zA-Z]|\\.[a-zA-Z.])[a-zA-Z0-9._]*";return {c: [ e.HCM,e.CBCM,{b:r,l:r,k:{keyword:"maximize subject to var ",literal:""},r:90},{cN:"number",b:"\\d+(?:\\.\\d*)?(?:[eE][+\\-]?\\d*)?i?\\b",r:0}]} });

	// console.log(hljs.listLanguages());
	document.querySelectorAll('pre code').forEach((block) => { hljs.highlightBlock(block); });
});
</script>

</body>
</html>

```

[...]

PR #3026, opened 2021-10-04
Add option for highlight plugin to register additional languages via callback

rajgoel:
This PR adds an optional parameter `callback` to the configuration of the `highlight` plugin. If a callback function is provided, the function is called with `hljs``as argument before the plugin starts to highlight the code.

# Sample configuration

```js
Reveal.initialize({
  // ...
  highlight: {
    callback: registerAMPL 
  },
  // ...
});
```
A new language can then be added, by including the respective function as shown below:
```js
function registerAMPL(hljs) {
  hljs.registerLanguage("ampl",function(hljs){
    return   {
      contains: [ 
        hljs.HASH_COMMENT_MODE,
        hljs.C_BLOCK_COMMENT_MODE,
        {
          className: "number",
          begin: String.raw`\b(\d*\.?\d+)\b`,
          relevance:0
        },
        {
          className: "strong",
          begin: String.raw`^ampl[\?:]`,
          relevance:0
        }
      ],
      keywords: { 
        keyword: 'maximize minimize subject to set within union inter diff symdiff cross sum in by mod var param default setof dimen symbolic if then else',
        type: 'binary integer',
        string: 'reset model solve data option solver solver_msg solution_precision display include print printf quit',
        built_in: 'abs max min sqrt round trunc ceil floor',
        number: 'Infinity',
        symbol: '_nvars _varname _var'
      } 
    };
  });
}
```

Fixes https://github.com/hakimel/reveal.js/issues/2761


rajgoel:
Thanks a lot for including this :+1:

maxmeyer:
This is very helpful. @rajgoel Is there any place where this is documented?

hakimel:
I added it to the docs here: https://revealjs.com/code/#the-highlight.js-api-%26-beforehighlight-4.2.0

Martinomagnifico:
Great addition, Asvin!

maxmeyer:
@hakimel  Great! Thanks.

PR #3075, opened 2021-11-21
fix npm start -- --host=0.0.0.0 unuseful

cashcat:
I start the server in my cloud server，but it just can run with localhost, even I use "npm start -- --host=0.0.0.0 --port=8001" ,it does anything.


hakimel:
Thanks!

Issue #2179, opened 2018-06-18
background videos not playing when hosted on webserver


supersambo:
Hi, 
I have a presentation that makes use of background videos. I just finished the presentation and everything works fine when run locally by just opening the .html file. However when uploading this to an apache2 webserver the background images won't work. 
I'm not sure if this is an issue with my webserver configuration (which is default of a fresh apache 2 installation) or if this has to do with reveal.js. itself. I'm using mp4 videos.

Best, 
stephan

bohdantrotsenko:
Back in days when I was using windows this was common as windows has case-insensitive filesystem.
Please check that your webserver does serve the video by downloading it and check the case sensitivity too.

boschkundendienst:
@supersambo and @bohdantrotsenko : Do you reference the video by full url like `http://...` or relative like `somefolder/somevideo.mp4`. Is the video hosted on an https or http webserver? There might be a problem with mixed content, when hosting the presentation and video on different sites. If one uses http and the other https. - Just an idea.

Issue #1027, opened 2014-10-08
data-background-video not working on chrome mobile (android)


vitorcarvalhoml:
Hello, the new option found in dev branch 'data-background-video' doesn't work on some android device using chrome.

On android 4.4 works when using the controls to move between slides.
On android 4.2.2 doesn't works  even when using controls to move in slideshow

How can I get more debug info about this?

demoneaux:
This works for me in Android. Closing for now until there are more details provided.

robertop87:
Reopen please @hakimel 
The issue persists.
Background videos does not play in Chrome/Firefox for Android
Sometime just plays the first video frame. This is possible disabling the transition effect, but in other case the background video does not play at all

Android Version: 4.4.4
Chrome version: any
RevealJs version: 3.4.1
Mobile: Samsung J5

PR #3078, opened 2021-11-23
Fixing background video not playing on certain small factor devices (android)

vanch3d:
There has been reports in the past of background video not playing properly on some devices. See https://github.com/hakimel/reveal.js/issues/1027 or https://github.com/hakimel/reveal.js/issues/2179. 

I've been facing the same issue recently when trying to play `Reveal.js` presentations on some IoT-type devices (android board, with WebView) and even from my phone (Samsung, Android 11, both Samsung Internet and Brave) 

Could not easily reproduce the problem because of also using embedded videos (full screen) that where properly working. 

Until I spotted a slight difference between the two video injection: the background video didn't have `type` added to the `source` tag generated from the `data-background-video` attribute in `Reveal.js`

The solution in this PR seems to be fixing the problem on all my devices. I cannot explain why it is working (or bugging in the first place), nor whether it will now work on every devices, including older versions of OS or browsers.

Also, worth noting that the solution in this PR is simply generating the MIME type out of the file extension, which is not a very reliable solution. But it will do, unless we want to make the specification of the  `data-background-video` attribute more complicated


vanch3d:
An alternative suggestion for a fix would be to have the MIME type included in the `data-background-video` using a secondary string delimiter, for example : 

```html
 <section
        data-background-video="http://example.com/video1.mp4:video/mp4,http://example.com/video2.webm:video/webm"
        data-background-video-loop
        data-background-video-muted
      >
        <h2>Background Video</h2>
</section>
```

This will still be backward compatible with the current handling of multiple sources and make the specification and validity of the MIME the responsibility of the designer, not the JS package.

hakimel:
Solving this for most cases without requiring additional authoring seems like a good solution for now. We can always add the ability to provide explicit types via new syntax later.

I update the implementation a little so that we only set known/valid MIME types instead of always using the file extension. c5d549810e7d329d200a4f30793a96d3d3b43614

hakimel:
Thanks for fixing this!

PR #3082, opened 2021-11-29
Add a dummy .map file to suppress "missing file" errors from some browsers.

dabrahams:
Especially as an inexperienced web developer, I need my presentations to function cleanly so that I get a clear indication when I've done something wrong.  If I just include reveal.js in my site without a map file, I see errors in the web development pane.


dabrahams:
@hakimel Hey, I am maintaining my own fork of reveal.js, solely to deal with this problem.  Is there another approach I should use?  I'd like to know if you think my approach is inappropriate, in which case we should close this issue.

Thanks

P.S. I am including reveal.js in my project as a Git submodule, so I can't just go in and add the file myself without having a repository from which to pull it.

hakimel:
Good catch—we shouldn't be referencing source maps that aren't available.

Instead of adding a dummy file, I've change it so the real source maps are included in the repo. I don't think that should cause any issues.

Issue #3080, opened 2021-11-26
Embedded presentations, two issues, single fix


Martinomagnifico:
* Issue 1: **In all browsers**, when we have any embedded presentation, remotely opening it in fullscreen won't let you use the keyboard immediately, only after another click. When the controls are set to false, as in some presentations, the user may not understand why it does not work. See the attached example of using remote triggers.
* Issue 2: **In Safari**, when we have any embedded presentation, selecting it and pressing 'F' will NOT scale the first slide initially, only after navigating. This is even visible in the main top presentation on revealjs.com, and in the "multiple-presentations.html" in the examples. 
* (Issue 3): **In Safari**, in some of the embedded of the examples, like on this page: https://revealjs.com/fullscreen/ and the multiple presentation example, fullscreen turns into a bar, not my idea of fullscreen. This is a combination of issue 2 and the flexbox markup.

Here's the fix:
Storing the `keyboardCondition` and temporarily setting it to null fixes issue 1 and 2. At first, using `layout()` on the deck inside the `fullscreenchange` function fixed issue 2 for me, but it seems that it is no longer needed when using the `keyboardCondition` fix of issue 1. Maybe changing the config automatically triggers a new layout, not sure about that. Issue 3 is fixed with some change to the markup/CSS compared to the original multiple presentations page.

Below is an example of multiple presentations that can be triggered by remote buttons. It includes the fixes above in the right-side presentation.  I added Screenfull to be able to assign the remote buttons to work, and in the FS coding only use some of the prefixes. I also made the .slides element somewhat clearer to show the Safari issue better.

```
<!DOCTYPE html>

<head>
    <meta charset="utf-8">
    <title>reveal.js - Multiple Presentations</title>
    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/reveal.js@4.2.0/dist/reveal.css">
    <link id="theme" rel="stylesheet" href="https://cdn.jsdelivr.net/npm/reveal.js@4.2.0/dist/theme/white.css">
    <style>
        html {
            width: 100%;
            height: 100%;
            margin: 0;
            position: relative;
            display: block;
        }

        body {
            background: #ddd;
            width: 100%;
            height: 100%;
            margin: 0;
            position: absolute;
            top: 0;
            left: 0;
            z-index: 1;
            display: block;
        }

        .row {
            display: flex;
[...]

PR #3083, opened 2021-11-30
Safari fix

Martinomagnifico:
This is a fix for the fullscreen functionality in Safari, when launched from an embedded presentation. As discussed here: https://github.com/hakimel/reveal.js/issues/3080. The timeout is needed because there is no promise at the moment that the deck is actually fullscreen. 
The second part is the styling on the 'multiple-presentations' page, which sets the 50vh on the parent, not the deck itself, because Safari will not go full-fullscreen when a height is set on it. This parent-styling should also be implemented on the revealjs.com pages.

It does not fix the 'remote button' issue (as discussed in that same issue), because that will need either an event listener on fullscreen changes (which is not built in (yet)), or the possibility to both focus the target deck AND simulate an F keypress at the same time.


hakimel:
This issue should be resolved now: https://github.com/hakimel/reveal.js/issues/3080#issuecomment-988671446

Took a slightly different approach to trigger this additional layout in response to `fullscreenchange` instead of inside the keyboard listener.

PR #3084, opened 2021-12-01
don't override an explicit width or height

dabrahams:
I placed an image in my slide using `style='position: absolute...'` and explicit `width` and `height` attributes but it didn't show up.  Turns out markdown wrapped it in a `<p>`, whose size was then zero… which then forced the image size to zero.

### DIFF
--- a/css/theme/template/theme.scss
+++ b/css/theme/template/theme.scss
@@ -93,8 +93,12 @@
 .reveal img,
 .reveal video,
 .reveal iframe {
-	max-width: 95%;
-	max-height: 95%;
+	&:not([width]) {
+		max-width: 95%;
+	}
+	&:not([height]) {
+		max-height: 95%;
+	}
 }
 .reveal strong,
 .reveal b {


dabrahams:
I'm not sure this is really what's needed, because it seems like width or height specified in non-absolute units like '%' or 'em' *should* probably be limited by minima/maxima.  I'm not a CSS expert, so maybe there's a better remedy out there.

Jwhisenhunt16 (approved):

jonatanlr98:
> Coloqué una imagen en mi diapositiva usando atributos `style='position: absolute...'`explícitos `width`y `height`pero no apareció. Resulta que Markdown lo envolvió en un `<p>`, cuyo tamaño era entonces cero... lo que luego forzó el tamaño de la imagen a cero.

jonatanlr98 (approved):

jonatanlr98 (approved):

Mitumaru (approved):

jonatanlr98 (approved):

jonatanlr98 (approved):

Mitumaru (approved):

jonatanlr98 (approved):

jonatanlr98 (approved):

adwokat (approved):

PR #3155, opened 2022-03-14
Update demo.html

qq1418381215:
这里之前写错了，箭头会返回到第一张ppt


hakimel:
Thanks for spotting that. That link was supposed to go to the slide one level down, I've updated that in a separate commit.

PR #3157, opened 2022-03-16
RevealMath.MathJax3: Allow the skipHtmlTags option to be overriden

eric-wieser:
The previous logic here was backwards, and did not allow the user to override `options` in the mathjax config structure.
This makes it match how the `startup` and `tex` fields are merged.


hakimel:
Thanks @eric-wieser!

Issue #3133, opened 2022-02-18
Incorrect slide numbering when printing to PDF


csiegl182:
The slide numbering is incorrect when printing to PDF. It starts with two and inserts and increments by two on a horizontal slide transition. Moreover, the numbering format (slideNumber property) is not evaluated when generating the PDF output.

chivongv:
What browser were you using when printing to PDF? According to [the docs on PDF export](https://revealjs.com/pdf-export/), the feature has only been confirmed to work in Google Chrome and Chromium.

csiegl182:
I am using Chrome. However, I think the issue is obvious.

Regarding the 4.3.1 tag:

In line 151 the index of the slide iteration is used for the slideNumber value. The index is not the right measuere since it increments by two on a horizontal slide change. I my personal fork I use a separate variable for slide number and increment it individually. This works fine.

Moreover, the slideNumber property needs to be considered somehow in the if-statement of line 150.

hakimel:
This should be fixed as of https://github.com/hakimel/reveal.js/pull/3182

We don't support custom formatting of page numbers when printing.

PR #3182, opened 2022-04-02
fix: pdf page numbering for fragment group

chivongv:
### What this PR does
It fixes the PDF page numbering when using reveal.js PDF export feature. Please check the attached pdf files below. Compare pages 3, 9 and 18. This might fix the issue [#3133](https://github.com/hakimel/reveal.js/issues/3133).

[demo - before.pdf](https://github.com/hakimel/reveal.js/files/8403043/demo.-.before.pdf)
[demo - after.pdf](https://github.com/hakimel/reveal.js/files/8403044/demo.-.after.pdf)

### Question
I used "." as the default separator for fragment group on line 188 in _js/controllers/print.js_. Should the config file accept a new config option for it instead?


hakimel:
This looks great, thanks for the PR!

Issue #2882, opened 2021-01-24
share video accross multiple slides or fragements w/o restarting the video


think-nice-things:
Hi,

I just found that sharing a background video accross multiple slides or fragments restarts the video.
Is there a way to not restart the video but just keep it running?

#### share video accross multiple slides

If switching from the first to the second slide, the video restarts from the beginning.
I'd rather like to keep it running and not restarting it.

```html
<section>
  <section data-background-video = "myvideo.mp4" data-background-video-loop> 
    content
  </section>
  <section data-background-video = "myvideo.mp4" data-background-video-loop> 
    content
  </section>
</section>
```

#### share video with fragments

If switching to the second fragment the video restarts from the beginning.
I'd rather like to keep it running and not restarting it.

```html
<section data-background-video="myvideo.mp4" data-background-video-loop>
  <div>
    <h2>Some caption</h2>
  </div>
  <div class = 'fragment'>
    Text which appears as fragment later.
  </div>
</section>
```

any ideas?

rben01:
I have what I believe is the same issue, but for ordinary videos, not background videos. When a `![](video.mp4)` on a slide is playing, every fragment on the slide will restart the video. I'd expect the video to just continue playing through the fragments.

hakimel:
Multiple identical & consecutive background videos will now keep playing without interruption as of https://github.com/hakimel/reveal.js/commit/a29a9c71aed1c34abd75cd32ad373d1af23fbb1c

I have not been able to reproduce the issue with fragments restarting videos so assuming that's already been fixed.

rben01:
> Multiple identical & consecutive background videos will now keep playing without interruption as of [a29a9c7](https://github.com/hakimel/reveal.js/commit/a29a9c71aed1c34abd75cd32ad373d1af23fbb1c)
> 
> I have not been able to reproduce the issue with fragments restarting videos so assuming that's already been fixed.

I don't think so, on master there is this: https://github.com/hakimel/reveal.js/blob/a29a9c71aed1c34abd75cd32ad373d1af23fbb1c/js/controllers/slidecontent.js#L310-L312

I believe this runs unconditionally when a fragment is advanced, and causes all videos on the slide to restart.

hakimel:
[...]

PR #3189, opened 2022-04-10
fix: video restarting on same bg

chivongv:
## What this PR does
This PR should fix [#2882](https://github.com/hakimel/reveal.js/issues/2882). There are some minor problems with this PR. Please check below.

## Minimal reproduce
Copy and paste the code below to _index.html_

```html
<div class="reveal">
	<div class="slides">
		<section>
			<section 
				data-background-video="https://www.w3schools.com/html/movie.mp4"
				data-background-video-loop
			>
				Some content
			</section>
			<section
				data-background-video="https://www.w3schools.com/html/movie.mp4"
				data-background-video-loop
			>
				Some additional content...
			</section>
		</section>
    </div>
</div>
```

## Problems

- [ ] 1. There is a small flicker/delay on the video when changing between two slides.
- [ ] 2. State are not syncing. Video on next slide will play regardless if previous video is paused or not.
- [ ] 3. Slides with video as background must be directly after each other or the video will restart e.g. <Slide 1 bg-video> <Slide 2 text> <Slide 3 bg-video>

If these problems are not so serious then this PR does what it is intended to do. Comments/suggestions are welcome.


Mitumaru (approved):

hakimel:
Thanks for the PR! I debugged this today was able to work around the background flicker by carrying forward the whole `<video>` element instead of only the current time. Available in master now: https://github.com/hakimel/reveal.js/commit/a29a9c71aed1c34abd75cd32ad373d1af23fbb1c

Issue #3120, opened 2022-02-03
Print PDF and class "r-fit-text"


danekt:
Class "r-fit-text" does not work in "?print-pdf" print preview mode and after saving as PDF file in Google Chrome browser.
E.g. <h1 class="r-fit-text">Very long first level heading text</h1> is truncated in print preview only to "Very long first le" :(

PR #3191, opened 2022-04-10
fix: truncated long r-fit-text on pdf

chivongv:
## What this PR does
This PR should fix [#3120](https://github.com/hakimel/reveal.js/issues/3120) by setting white-space to normal. White-space is being set to nowrap in _css/layout.scss_ on line 22. Read about white-space [here](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space) and compare between _nowrap_ and _normal_ option.

- [before.pdf](https://github.com/hakimel/reveal.js/files/8458697/before.pdf)
- [after.pdf](https://github.com/hakimel/reveal.js/files/8458696/after.pdf)

## Minimal reproduce
Copy and paste the code below in index.html

```html
<div class="reveal">
      <div class="slides">
        <section>
          <p>Add the <code>r-fit-text</code> class to auto-size text</p>
          <h1 class="r-fit-text">
            Very long first level heading text text text text text
          </h1>
          <p>Press <strong>ESC</strong> to enter the slide overview.</p>
          <p>
            Hold down the <strong>alt</strong> key (<strong>ctrl</strong> in
            Linux) and click on any element to zoom towards it using
            <a href="http://lab.hakim.se/zoom-js">zoom.js</a>. Click again to
            zoom back out.
          </p>
          <p>(NOTE: Use ctrl + click in Linux.)</p>
        </section>
      </div>
</div>
```


Mitumaru (approved):

hakimel:
Thanks for the PR.

I merged this, but as I was testing I noticed other cases where fit-text was overflowing in PDFs. I fixed those problems by invoking the fit-text script after all pages have been added to the DOM so it gets the right measurements. That fix also solved the issue in this PR so I ended up removing your change to the PDF styles.

Fitty—the library we're using to fit text—will automatically handle line wrapping in PDFs now that it works correctly.

chivongv:
Oh, I see. Fitty seems to be a better solution. Thanks for letting me know!

Issue #1529, opened 2016-03-16
Zoom and speaker notes window


ced42:
Hello,

When two windows are opened (presentation and speaker notes), zoom function on any element works only on one window. I'm wondering if it a bug.

Regards,
Cédric.

obilodeau:
It would be nice to have the ability to zoom from the speaker notes view and have it synchronized with the presentation view.

janisdd:
Unfortunately this framework moved away from zooming to elements but now zooms to mouse position...

I just recently needed this feature and implemented it for zooming to elements (I use an older version).
I'm sure I can adapt this to zoom to mouse position.

However, there is the issue with different sizes/aspect ratios of the main and speaker view.

I'll create a PR

PR #3212, opened 2022-05-21
fixes issue #1529 - syc main and speaker view zoom

janisdd:
Fixes issue #1529 - syc main and speaker view zoom

It uses a BroadcastChannel to communicate between the windows. The sync is in both ways.

The new config option `syncZoom` can be used to disable the sync.

I also noticed that `zoomLevel` is used anywhere outside this plugin...

I could not find anything about semicolons in `CONTRIBUTING.md`, so I just tried to stick with the general style and add them (I normally leave them out)

### DIFF
--- a/js/config.js
+++ b/js/config.js
@@ -275,6 +275,9 @@ export default {
 	// viewDistance in order to save resources.
 	mobileViewDistance: 2,
 
+	// if speak view and the normal view should sync the zoom
+	syncZoom: true,
+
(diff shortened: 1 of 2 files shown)


janisdd:
Well one thing I didn't thought about (until now) is that every browser tab with this url/presentation will be synced... If this is a problem we need to add a guid to distinguish the tabs (maybe there is such a thing already in reveal.js?).

PR #3257, opened 2022-08-12
Missing lang attribute

sojinsamuel:
The lang attribute does not default to English. It defaults to an unknown, which is an accessibility issue.


hakimel:
Thanks!

PR #3291, opened 2022-09-29
Fix overview spacing for disabled auto layout

Proliecan:
When `disableLayout:true` is set in the main html of the presentation and the viewpoint was higher or wider than the configured slide size,
slides used to overlap in the overview page.

The changes made to `js/reveal.js` should fix that.

comment:
I ran `gulp js` afterwards.


Proliecan (commented):
Double checked my changes.  
Good to go for MR

Proliecan:
@hakimel

hakimel:
Thanks!

PR #3294, opened 2022-10-06
assign --r-heading5 to r.eveal h5 in scss

signedav:
The size that is defined for --r-heading5-size is not taken into account because it was not set in the templates.

Feel free to close this PR if my change does not make sense.


hakimel:
Thanks for your PR. There is no SCSS variable named `$heading5Size` so unfortunately this would not work.

PR #3304, opened 2022-10-22
Plugin to allow slider comparison of images

oliche:
Adapted a plugin from the work here:
https://github.com/CreativeTechGuy/RevealJS

This is useful to present comparison of images at the same scale and makes for a great widget in scientific presentations.


Martinomagnifico:
Hi Olivier, it is not a plugin. Merging it to the main branch makes no sense. Furthermore, it requires multiple `.reveal` containers if the user wants more “image comparisons”, while a single presentation only has a single `.reveal` wrapper.

PR #3305, opened 2022-10-24
Gulp livereload: include subfolders to watch for changes in html and md

lolmaus:
Currently it's only watching for changes to `.html` and `.md` files located in the root of the project. I was frustrated when livereload stopped working for me: turns out it was because I put my content into subfolders.

The new config works for files both in root and in subfolders.


hakimel:
Makes sense. Thanks!

PR #3310, opened 2022-10-31
Black & White compact themes with verbatim headers.

peter-kehl:
Dear Hakim,

Thank you for Reveal.js, and for your continuous support and improvements.

This pull request adds two themes: `black_contrast_compact_verbatim_headers.scss` and `white_contrast_compact_verbatim_headers.scss`. They're based on your `black.scss` and `white.scss`, with modifications
- headers are verbatim (rather than in CAPITAL CASE) - useful when the words/terms are in CamelCase or mixed;
- compact (less spacing) - using the screen more;
- high contrast (pure black on pure white, and vice-versa).

This would help my existing presentation (https://peter-kehl.github.io/no_std_rust_libs = https://github.com/peter-kehl/no_std_rust_libs) and 18+ more presentations to come (https://github.com/peter-kehl/no_std_patterns).

Could there be a better way to organize/centralize the code shared between `black.scss` and `black_contrast_compact_verbatim_headers.scss` (and between `white.scss` and `white_contrast_compact_verbatim_headers.scss`), please do as you see fit.

Thank you in advance either way. Related to #3312.

### DIFF
--- a/css/theme/source/black_contrast_compact_verbatim_headers.scss
+++ b/css/theme/source/black_contrast_compact_verbatim_headers.scss
@@ -1,0 +1,49 @@
+/**
+ * Black compact & high contrast reveal.js theme, with headers not in capitals.
+ *
+ * By Peter Kehl. Based on black.(s)css by Hakim El Hattab, http://hakim.se
+ *
+ * - Keep the source similar to black.css - for easy comparison.
+ * - $mainFontSize controls code blocks, too (although under some ratio).
+ */
+
+
+// Default mixins and settings -----------------
+@import "../template/mixins";
+@import "../template/settings";
+// ---------------------------------------------
+
+
+// Include theme-specific fonts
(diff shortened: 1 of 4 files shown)


peter-kehl:
#3312

hakimel:
Thanks for taking the time to create this, Peter!

I like the idea of a max contrast theme and have merge this and renamed them to black-contrast/white-contrast. However I undid the changes you made to font size and text transform because I don't think it scale very well to provide individual themes for smaller variations like that. It was hard to think of a good reason to have those properties vary between the black vs black-contrast themes when in all other aspects they are so similar.

To achieve the same style as in your original theme, you can use the black/white-contrast themes and append the following style tag anywhere on your page:

```
<style>
  :root {
    --r-main-font-size: 25px;
    --r-heading-font-weight: 450;
    --r-heading-text-transform: none;
  }
</style>
```

Issue #3315, opened 2022-11-05
background-image URLs containing parentheses fail to load


waldenn:
Hi. I'm seeing that background-image URLs containing any parentheses are not being displayed.

Example:
- [Working example](https://codepen.io/i_cant_rap/pen/GqoYBw)
- [Failing example](https://codepen.io/ccl123/pen/eYKdmKY)
- [Another failing example](https://codepen.io/ccl123/pen/abKmzXm)

The failing image URL is: https://commons.m.wikimedia.org/wiki/Special:FilePath/Apatosaurus%20skeleton%20construction%20(3349699758).jpg?width=600px (and loads fine as an image in the browser or directly)

I use revealjs on my conze.pt website, where I use Wikimedia Commons images, which often contain these parentheses in the URL.

PR #3316, opened 2022-11-05
escape parentheses in background image URL

waldenn:
This fixes the issue I posted earlier today: https://github.com/hakimel/reveal.js/issues/3315


waldenn:
There could be some other characters which need to be encoded (single/double quotes?)

Martinomagnifico:
Thanks for the PR. Your mention that there could be other characters led to the correct fix that was pushed earlier today.

Issue #3207, opened 2022-05-11
Code block lines wrongly spaced


gabyx:
I have

```html
    <section data-markdown>
        <script type="text/template">
        # Whats in it?

        <pre><code data-noescape data-line-numbers="|4|25|21|6|18-20|11|47-57" data-fragment-index="1">cmake-general/
  ├── a
      └── windows
              ├── a
              ├── b
              └── c
        </code></pre>
        </script>
    </section>
```
which renders as:
![image](https://user-images.githubusercontent.com/647437/167842019-833404f1-7e7b-450d-ab1c-a8c55dabb99a.png)

What is wrong here?

PR #3323, opened 2022-11-19
fix(plugin/markdown): replace leading spaces at start of line

xhalo32:
Fixes various bugs with incorrect markdown formatting


xhalo32:
I believe fixes #3207

azat:
Any news on this one?

azat:
Although forget it, I've just tested this PR, and breaks lots of Markdown slides.

Issue #2785, opened 2020-09-28
Internal web server should listen on 127.0.0.1, not 0.0.0.0


dnnr:
It seems that a commit from March (5a5a5c9a) inadvertently changed the default listening interface from localhost to 0.0.0.0, meaning every network interface available.

This is an unsafe default and a potential security risk (information disclosure), especially since there appears to be no obvious/documented way to change this behavior, except for modifying the source.

Please consider making 127.0.0.1 the default again. Thanks.

cashcat:
but how to start with 0.0.0.0?

hakimel:
As of #3075 you can use:
```
npm start -- --host=0.0.0.0
```

Thanks @cashcat for fixing that.

PR #3357, opened 2023-01-25
change binding host from localhost to 0.0.0.0

RapTho:
## Problem
The binding host of 'localhost' is an issue when containerizing your final reveal.js app. The presentation will not be accesible from outside the container and thus port mapping from host to container port doesn't return the desired result.
The request returns 'ERR_EMPTY_RESPONSE'

## Steps to reproduce
Try to build a container with your binding host of 'localhost'. 
`docker build -t myreveal:1.0 .`

Example Dockerfile: https://github.com/RapTho/selfIntroductionRevealJS/blob/main/Dockerfile

Run the container:
```
docker run -d \
          --name revealjs \
          -p 8000:8000 \
          --restart unless-stopped \
          myreveal:1.0
```

access the browser at http://localhost:8000

## Solution
Change the binding host from 'localhost' to '0.0.0.0' as proposed in this PR

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -24,7 +24,7 @@ const autoprefixer = require('gulp-autoprefixer')
 
 const root = yargs.argv.root || '.'
 const port = yargs.argv.port || 8000
-const host = yargs.argv.host || 'localhost'
+const host = yargs.argv.host || '0.0.0.0'
 
 const banner = `/*!
 * reveal.js ${pkg.version}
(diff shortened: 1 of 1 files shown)


hakimel:
We originally defaulted to 0.0.0.0 but switched to localhost for the reasoning in #2785

If you want to use 0.0.0.0 as host you can start reveal.js like this:
```
npm start -- --host=0.0.0.0
```

PR #3358, opened 2023-02-02
Correctly strip leading white-space from markdown

azat:
*Note, this is resubmit of #1831 (cc: @jerrykan)*

If the markdown contains something that is indented by more that the `leadingTabs`/`leadingWs` then extra white space is incorrectly removed. ie the following example:

```
    <section data-markdown>
    some text
       indented text
          more indented text
    </section>
```

would result in the following markdown:

```
some text
   indented text
  more indented text
```

We can work around this problem by using a function to generate the replace value.


azat:
Any news on this one?

@Martinomagnifico maybe you can take a look? (Since you closed #3323, and apparently you are familiar with the code base).

Thanks in advance.

azat:
Understand, thanks. It will be great if you could ping someone else who had knowledge about Markdown (or maybe I should use `git blame` and ping by myself, or maybe I just need to wait more time?)

azat:
Friendly ping on this one, maybe this will bring someones attention...

hakimel:
Thanks @jerrykan @azat — I tried this out and the fix seems to be working very well.

PR #3367, opened 2023-02-21
Fixes #3366 Add a 'real'-black theme

### LINKED ISSUE #3366: Can we have a 'real'-black theme ?
The name suggests black, however the background-color is not actually black:

https://github.com/hakimel/reveal.js/blob/7de6ccb65b98f657f757cb5610613ad9d4f3da2b/css/theme/source/black.scss#L19

for OLED displays that can do nice black this is 'wasteful' ;-)

gbraad:
Minor modification to change background to black

### DIFF
--- a/css/theme/source/real-black.scss
+++ b/css/theme/source/real-black.scss
@@ -1,0 +1,47 @@
+/**
+ * Real black theme for reveal.js. This is the opposite of the 'white' theme.
+ *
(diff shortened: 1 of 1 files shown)


hakimel:
There was another PR open with a similar change that I just merged #3310

If you pull the latest master, the new black-contrast theme provides you with a fully black background :)

PR #3395, opened 2023-04-13
Add link to examples and revealjs.com to index.html slide

timabell:
Took me a little while to find the examples, so this makes them a bit more discoverable.

Thanks for software, very useful.

## Before

shutter_20234-13_04

## After

shutter_20234-13_03

### DIFF
--- a/examples/build-index.sh
+++ b/examples/build-index.sh
@@ -1,0 +1,10 @@
+#!/bin/sh
+
+rm index.html
+
+echo "<h1>reveal.js example decks</h1><ul>" >> index.html
+for deck in `ls *.html`; do
+    name="${deck%.html}"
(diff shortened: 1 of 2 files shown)


t-fritsch:
thanks for the PR @timabell ! 🙏 

Making examples more visible is a good idea, maybe adding this to the README would be better (in the list of links in [getting started section](https://github.com/hakimel/reveal.js#getting-started)) ? 

I guess @hakimel leaves default slides empty (apart "Slide 1" and "Slide 2" titles) to keep the default deck really simple, but maybe having this in the first slide is fine tho ?

Regarding the `examples/index.html` and `examples/build-index.sh` I wonder if it is really needed since we already have a file listing in the defaut webserver (launched with `npm start`) : 
image

eudeq124:
Ok c’est compris tu me laisse ton whatssap

PR #3435, opened 2023-07-18
fix(package.json): upgrade a minor version to handle the last fix published

RomanRyckebusch:
This fix doesn't upgrade the package version: https://github.com/hakimel/reveal.js/commit/0b44308754d69d2cf596915b6eb226104718d096

The patch is not released because the package is stuck at version 4.5.0 in the NPM registry.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "reveal.js",
-  "version": "4.5.0",
+  "version": "4.5.1",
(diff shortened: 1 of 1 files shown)


t-fritsch:
thanks for the PR 👍  I'm closing it cause I think this isn't relevant anymore (many new versions have been released since) :)

PR #3441, opened 2023-08-03
fixes livereload when using root CLI param

t-fritsch:
when loading an index.html file outside the reveal folder like this 
```bash
npm start --root=../some/path
```
livereload is broken since gulp only looks inside the reveal folder.

This fix allows to make livereload work again even when using slides in another folder.


hakimel:
Makes sense, thanks!

PR #3443, opened 2023-08-04
adds ability to override markdown plugin default options

t-fritsch:
This PR allows user to override markdown plugin default values for every options (`separator`, `notesSeparator` and `verticalSeparator`).

I've made many slides with reveal and was always ending copy/pasting the same attributes on every `<section data-markdown ...>`, giving something like this : 

```html
<section data-markdown="00-intro.md" 
    data-separator-vertical="\n--\n" 
    data-separator-notes="\n~~\n"></section>
<section data-markdown="01-ecmascript.md" 
    data-separator-vertical="\n--\n" 
    data-separator-notes="\n~~\n"></section>
<section data-markdown="02-react.md" 
    data-separator-vertical="\n--\n" 
    data-separator-notes="\n~~\n"></section>
...
```

With this PR we can now remove every repeating html attributes : 
```html
<section data-markdown="00-intro.md"></section>
<section data-markdown="01-ecmascript.md"></section>
<section data-markdown="02-react.md"></section>
...
```

and simply pass options to markdown plugin like this : 

```js
Reveal.initialize({
    markdown: {
        defaultOptions: {
            verticalSeparator: '\n--\n',
            notesSeparator: '\n~~\n',
        }
    }
})
```

If there is a better way to achieve the same result don't hesitate to tell me 👍


hakimel:
Being able to provide defaults for those options is great. I made a small change to this and moved the new separator options out of `defaultOptions` so to set them you'll want to use:
```js
Reveal.initialize({
    markdown: {
        verticalSeparator: '\n--\n',
        notesSeparator: '\n~~\n',
    }
})
```

t-fritsch:
yes it's much simpler like that 👍  

Thanks a lot for the other PR merges by the way !

PR #3452, opened 2023-08-22
fix the "go down"

guofei9987:
href="./1" to go down. If use href="#", it will go to the first page.


hakimel:
Hey, this button uses a special class (`.navigate-down`) which tells reveal.js to monitor it to for clicks and trigger a downward navigation. That said there's no need for any special href target on it.

PR #3453, opened 2023-08-22
fix typos in variable names

individual-it:
while looking into the plugin I've seen these small typos and fixed them to make it easier to work with the code


hakimel:
Thanks for fixing!

individual-it:
wow, this was probably the fastest PR ever merged!

We are currently working on a small project to add more functionality to the markdown plugin, so more PRs will follow from the https://github.com/JankariTech team

PR #3454, opened 2023-08-22
Refactored var to let or const, used strict equality operator

grgprarup:
While working with the markdown plugin, I've seen some semantic errors so fixed them. The changes are done in markdown plugin only.
The changes made in this PR:
- Refactored to replace `var` declarations with `let` or `const` for better variable management
- Refactored the code to replace `==` with `===` and `!=` with `!==`

### DIFF
--- a/plugin/markdown/plugin.js
+++ b/plugin/markdown/plugin.js
@@ -38,15 +38,15 @@ const Plugin = () => {
 	function getMarkdownFromSlide( section ) {
 
 		// look for a <script> or <textarea data-template> wrapper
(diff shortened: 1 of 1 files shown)


hakimel:
Thanks!

t-fritsch:
great work ! 
@hakimel shouldn't the `plugin/markdown/markdown.js` and `plugin/markdown/markdown.esm.js` files get updated too (or maybe it is part of a github action/git hook somewhere) ?

hakimel:
@t-fritsch Thanks for the reminder, I just recompiled them.

t-fritsch:
you're welcome ! I wasn't sure this was the way 😅

PR #3456, opened 2023-08-24
Update demo.html

menotti:
switch themes keep current slide


hakimel:
Nice enhancement, thanks!

PR #3457, opened 2023-08-24
fix code blocks font when printing pdf

### LINKED ISSUE #2867: Print-pdf changes code font back to Courier
When performing a print-pdf of my slides, slides with any code snippet where I have a custom CSS with Source Code Pro, the Source Code Pro gets overridden with Courier.  It's minor but would like to maintain the same aesthetic in both slides and pdf. Thanks

Reveal.js in the browser
Screen Shot 2020-12-17 at 8 41 51 AM

Reveal.js with print-pdf
Screen Shot 2020-12-17 at 8 42 30 AM

t-fritsch:
fixes #2867

looks like the font-family rule was no more useful, with this PR code blocks should now use the default `--r-code-font` value of the current theme (no more override).


hakimel:
🙌

PR #3461, opened 2023-08-26
Update layout.scss

Vivekv704:
Practice


t-fritsch:
I think you made a mistake here @Vivekv704 :)

PR #3464, opened 2023-09-02
fix dracula li markers

t-fritsch:
when working on my presentation I noticed that `dracula.css` was modified because sass file hasn't been altered in #3450 (only the built css file has been edited)

At first this PR aimed only to simply apply the same patch to the sass file to make it match the compiled version, but then I noticed that dracula's way to render `<li>` marker was by adding `:before` pseudo elements resulting in always the same bullet point, whatever nesting level it was into (always a disc). In other themes, the li renders differently depending on the nesting level (disc, square, circle).

By replacing `:before` pseudo elements with `::marker` selector, we can have the same style as in other themes :
dracula-list-item-comparison

from left to right : before this PR / PR / original "dark" theme for comparison

**NB:**  The bullets are now bigger than before this PR, but it is consistent with other themes. If one prefers smaller markers, it can be done in custom CSS : 
```css
.reveal ul li::marker {
    font-size: 70%; /* or whatever scale fits */
}
```

### DIFF
--- a/css/theme/source/dracula.scss
+++ b/css/theme/source/dracula.scss
@@ -86,20 +86,17 @@ $codeFont: "Fira Code", $systemFontsMono;
 	--r-list-bullet-color: #{$listBulletColor};
 }
 
-.reveal strong, .reveal b {
-	color: var(--r-bold-color);
-}
-
-.reveal em, .reveal i, .reveal blockquote {
-	color: var(--r-italic-color);
-}
-
-.reveal code {
-	color: var(--r-inline-code-color);
-}
-
-// Dracula colored list bullets and numbers
 .reveal {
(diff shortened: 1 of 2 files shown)


t-fritsch:
Just noticed there is a regression on `ol` numbering. Will fix before marking ready for review.

t-fritsch:
fixed ol numbering 
dracula-fix-li-numbering

did some sass syntax optimization too (remove duplicate `.reveal` selectors)

hakimel:
Oops, completely missed that that #3450 only changed the source files.

Thanks for catching that and for making the styles consistent with other themes.

t-fritsch:
my pleasure !

PR #3468, opened 2023-09-08
fix video restarting on same bg #2882

slavaGanzin:
Fixes: https://github.com/hakimel/reveal.js/issues/2882
Alternative to: https://github.com/hakimel/reveal.js/pull/3189

Before:

https://github.com/hakimel/reveal.js/assets/1011721/42fe8476-4ab1-4e26-9ef3-c6c254ba2178

After:

https://github.com/hakimel/reveal.js/assets/1011721/6b94450b-74f2-4fe7-916d-22431de0e6c0

This fixes only transition between fragments, between slides it's more complicated and should be handled by somebody with deep knowledge of internals like @hakimel himself :)

### DIFF
--- a/js/controllers/backgrounds.js
+++ b/js/controllers/backgrounds.js
@@ -281,43 +281,47 @@ export default class Backgrounds {
 
 		} );
 
+    let previousBackgroundHash = this.previousBackground ? this.previousBackground.getAttribute( 'data-background-hash' ) : null;
+    let currentBackgroundHash = currentBackground.getAttribute( 'data-background-hash' );
+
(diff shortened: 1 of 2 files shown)


hakimel:
Hey! Support for continuous background video playback across slides was added in https://github.com/hakimel/reveal.js/commit/a29a9c71aed1c34abd75cd32ad373d1af23fbb1c

I've been testing showing and hiding fragments but haven't seen them have any effect on inline or background video playback. Can you check if you're still experiencing this in the latest version? If you are, I'd really appreciate an example presentation where this can be reproduced. Appreciate it!

PR #3477, opened 2023-09-17
Notes plugin: notes from data-notes attribute were not shown

### LINKED ISSUE #3476: Notes plugin: data-note does not work on slide-level
We have 2 different ways to specify notes: with `data-notes` attribute and also with separate tag `<aside class="notes"`. But former does not work for a slide(but works at fragment level btw)

Example:

```html
<section>
  notes will be shown
  <aside class="notes">slide-level notes</aside>
</section>
<section data-notes="slide-level notes">
  notes will NOT be shown
</section>
```

skyboyer:
Fixes #3476 : `data-notes` worked on fragment level, but not at slide level

### DIFF
--- a/plugin/notes/plugin.js
+++ b/plugin/notes/plugin.js
@@ -146,7 +146,7 @@ const Plugin = () => {
 		}
 
 		// Look for notes defined in an aside element
-		if( notesElements ) {
+		if( notesElements && notesElements.length ) {
 			messageData.notes = Array.from(notesElements).map( notesElement => notesElement.innerHTML ).join( '\n' );
 			messageData.markdown = notesElements[0] && typeof notesElements[0].getAttribute( 'data-markdown' ) === 'string';
 		}


plugin/notes/plugin.js
```diff
@@ -146,7 +146,7 @@ const Plugin = () => {
 		}
 
 		// Look for notes defined in an aside element
-		if( notesElements ) {
+		if( notesElements && notesElements.length ) {
```
skyboyer, line 149: `notesElements` is either `null`(if there is `.current-fragment`) or a list of elements(result of `slideElement.querySelectorAll( 'aside.notes' )`). 

So if we have a slide:
1. without fragments at all(so no `.current-fragment`)
2. with notes specified through `data-notes`

We did never see the notes because condition `if( notesElements )` was always `true` even though list was empty

hakimel:
Good catch. Thanks for the fix.

PR #3489, opened 2023-10-06
Support to 'wheel' event listener

quochuy:
For some reasons the current implementation didn't work on my machine (Mac OS X + Chrome) despite having enabled mousewheel in the config. After some tests, adding a listener for "wheel" event fixed it.


hakimel:
Thanks! I went ahead and removed the deprecated listeners too.

quochuy:
That's great thanks.
What's new in 5.0.0? Can we directly update 4.x.x to it?

hakimel:
The only breaking change is the new scroll view being automatically enabled on mobile devices. There's a one-line config to turn that off though.

https://github.com/hakimel/reveal.js/releases/tag/5.0.0

PR #3503, opened 2023-10-27
Update README.md

GitProHub11:
typo fixed

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,7 +9,7 @@
 
 reveal.js is an open source HTML presentation framework. It enables anyone with a web browser to create beautiful presentations for free. Check out the live demo at [revealjs.com](https://revealjs.com/).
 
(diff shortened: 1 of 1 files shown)


t-fritsch:
thanks for the PR ! 
why do you think every item in this list should be with a capitalized first letter ?
It makes sense for Markdown, LaTex, and PDF (since their names are with uppercase first letter), but every other item in this sentence should stay lowercase imo 

I wondered if Auto-Animate should be lower-case or not but it's written this way in the documentation https://revealjs.com/auto-animate/ : 
> _All you need to do is add data-auto-animate to two adjacent slide <section> elements and_ **Auto-Animate** _will animate all matching elements between the two._

hakimel:
Thanks for the suggestion but no need to change this

PR #3507, opened 2023-10-28
chore: bump deps

Mister-Hope:
Note: marked is not bumped as marked has breaking changes across v4 and v9.

A separated PR will be sent later.


hakimel:
Thanks for taking the time to update those!

PR #3548, opened 2023-12-15
fix: use `setAttribute` instead of `innerHTML` to prevent XSS

### LINKED ISSUE #3546: Potential XSS vulnerability with `data-background-video` attribute
Hi @hakimel 

Thanks for all your hard work! 

we've encountered a potential security issue related to XSS (Cross-Site Scripting) when using the `data-background-video` attribute in slides. It appears that the way Reveal.js handles this attribute might allow for the execution of injected scripts, even when the attribute value is **HTML entity-encoded**.

Consider the following example where we use an encoded string in the `data-background-video` attribute:

```html
<section data-background-video="&quot;&gt;&lt;/video&gt;&lt;iframe srcdoc=&quot;&lt;script src='https://www.google.com/complete/search?client=chrome&q=123&jsonp=alert(document.domain)//'&gt;&lt;/script&gt;&quot;&gt;&lt;/iframe&gt;&lt;img src=wtf:&gt;&lt;a data-x=&quot;">
	<h2>🍦</h2>
</section>
```

The script will be executed despite being encoded. because `getAttribute` will [decode entities](https://stackoverflow.com/a/11224418) even if they are already escaped:
https://github.com/hakimel/reveal.js/blob/c23964274c17225d94be104289fa918eff1fbbd4/js/controllers/slidecontent.js#L102

And then concat in `innerHTML` will cause XSS:
https://github.com/hakimel/reveal.js/blob/c23964274c17225d94be104289fa918eff1fbbd4/js/controllers/slidecontent.js#L144-L152

---

Fix is pretty easy I think. Just replace `innerHTML` with attribute assignment, as shown below:

```js
backgroundVideo.split( ',' ).forEach( source => {
    let type = getMimeTypeFromFile(source);
    let sourceElement = document.createElement('source');
    sourceElement.setAttribute('src', source);
    if( type ) {
        sourceElement.setAttribute('type', type);
    }

    video.appendChild(sourceElement);
});
```

Using `setAttribute` ensures that the source remains a plain string.

If this approach seems suitable, I would be happy to create a PR to address this. Thank you for your consideration!

EastSun5566:
> Fixed #3546 

Hi @hakimel 

Following our previous discussion, I have made changes to use `setAttribute` instead of `innerHTML` to prevent XSS vulnerabilities.

I have created two small demos for you to try. The first demonstrates an XSS issue caused by using `innerHTML`:

https://jsfiddle.net/EastSun5566/9cfxeLzv/14/

The second uses `setAttribute` for attribute assignment:

https://jsfiddle.net/EastSun5566/rbogezv1/8/

You will notice that in the first case, an alert is triggered, while in the second, it is not. Although I believe URLs should be sanitized before setting attributes, I think this solution is good enough now.

Please let me know your thoughts. Thank you!

### DIFF
--- a/js/controllers/slidecontent.js
+++ b/js/controllers/slidecontent.js
@@ -142,13 +142,15 @@ export default class SlideContent {
 
 					// Support comma separated lists of video sources
 					backgroundVideo.split( ',' ).forEach( source => {
+						const sourceElement = document.createElement( 'source' );
+						sourceElement.setAttribute( 'src', source );
+
 						let type = getMimeTypeFromFile( source );
 						if( type ) {
-							video.innerHTML += `<source src="${source}" type="${type}">`;
+							sourceElement.setAttribute( 'type', type );
 						}
-						else {
-							video.innerHTML += `<source src="${source}">`;
-						}
+
+						video.appendChild( sourceElement );
 					} );
 
 					backgroundContent.appendChild( video );


hakimel:
Tested and seems to be working well, thanks!

PR #3563, opened 2024-01-28
feat: add a cssSelector configuration parameter RevealHighlight

TendTo:
This functionality is inspired by the [highlight.js](https://highlightjs.org/) library'[configuration](https://highlightjs.readthedocs.io/en/latest/api.html#configure).

While the standard "pre-code" selector works fine in most cases, there may be instances where the user may want more control over what elements the plugin will affect. This solution provides a very easy way to be more selective.

The documentation would need to be updated to account for this new option.


plugin/highlight/plugin.js
```diff
@@ -71,15 +72,15 @@ const Plugin = {
...
+			Array.from( reveal.getRevealElement().querySelectorAll( config.cssSelector ) ).forEach( block => {
 				Plugin.highlightBlock( block );
 			} );
 		}
 
 		// If we're printing to PDF, scroll the code highlights of
 		// all blocks in the deck into view at once
 		reveal.on( 'pdf-ready', function() {
-			[].slice.call( reveal.getRevealElement().querySelectorAll( 'pre code[data-line-numbers].current-fragment' ) ).forEach( function( block ) {
+			[].slice.call( reveal.getRevealElement().querySelectorAll( config.cssSelector + '[data-line-numbers].current-fragment' ) ).forEach( function( block ) {
```
TendTo, line 83: This line makes sense only for compatible modifications to the selector.
If you wanted to give complete control to the user, a new configuration would need to be added. Something like
```js
config.cssSelectorLineNumbers = typeof config.cssSelectorLineNumbers === 'string' ? config.cssSelectorLineNumbers : 'pre code[data-line-numbers].current-fragment';
```

PR #3568, opened 2024-02-05
fix vertical stack background not working in scroll view

bouzidanas:
Setting `data-background-color` on the parent `<section>` of a vertical stack works in normal slide view but not in the new scroll view ( `view: "scroll"` ).

In the following example,
```html
<div className="reveal">
    <div className="slides">
        <section data-background-color="red">
            <section> Vertical slide 1 </section>
            <section> Vertical slide 2 </section>
        </section>
        <section> Next horizontal slide </section>
    </div>
</div>
```
the first two slides which are in the same vertical stack should have the same red background. In the normal view they do, but in scroll view they do not.

Looking in `scrollview.js`, I could not find any code that aims to transfer the effect of the `<div class="backgrounds">` element which is responsible for establishing the background for stacks and slides so I added logic that checks if the "backgrounds" element exists and then copies the backgrounds of the immediate children to the corresponding scroll view page elements (in place of copying the presentation viewport background).

This seems to have resolved this issue for me.


hakimel:
Thanks! This has been merged with a few minor naming tweaks.

bouzidanas:
Hi @hakimel,

Your changes makes sense and target the particular issue more precisely with the new isVertical argument. However, I now have a theme background copy issue on singular slides! So the copying of the theme background happens for the stacks but not for a stack of 1? I see that a background element is created for the single slide but it is empty as far as I can tell.

It could be something else going on, but when I revert back to my fork, the issue is gone. I will look into it more a bit now and later when I am able.

bouzidanas:
Ok, I think I figured it out. By copying the presentation background and not the `slide-background` div background corresponding to the single horizontal slide, we omit the background styles added by the CSS rule with the selector `.reveal .slide-background` which should apply to every slide I think.

So right now, I am leaning on this being a bug. I thought it was because I was lazy loading my theme, but then I didnt understand why the code in my fork worked.

Issue #2645, opened 2020-04-08
? key will toggleHelp no matter what


pautasso:
_How to reproduce:_
If there is an <input> element on any slide, typing ? will show and hide the Help overlay message.

_How to fix:_ 
For some reason ? is handled in the `keypress` event, while the correct behavior to filter keyboard events in case there is an active input element is only found in the `keydown` event handler.

motyar:
Use `Reveal.configure({"keyboard":false})`

kno10:
The `?` key is implemented different from the other keyboard shortcuts.
It will toggle even in content editable elements, such as a text box of an interactive element.

Also, it uses charCode, which is [deprecated API](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/charCode) and additionally checks for the shift button.

https://github.com/hakimel/reveal.js/blob/aa5c03c234f198e86e32e198f44b07d10edcc78c/js/controllers/keyboard.js#L143-L150

Any reason to not remove this, and instead handle the help shortcut the same way as the other shortcuts in onDocumentKeyDown just below?

At the minimum, a guard like this would make sense IMHO:

https://github.com/hakimel/reveal.js/blob/aa5c03c234f198e86e32e198f44b07d10edcc78c/js/controllers/keyboard.js#L182-L184
https://github.com/hakimel/reveal.js/blob/aa5c03c234f198e86e32e198f44b07d10edcc78c/js/controllers/keyboard.js#L193-L195

hakimel:
Fixed. Not sure why this was originally in a `keypress` handler but I've moved it to `keydown` with all of the same safeguards as the other shortcuts.

gchriz:
Unfortunately this change breaks the toggleHelp on keyboards with a non-english layout. The "?" doesn't have keyCode 191 there. 

Since it's probably impossible to catch all international keyboard layouts regarding  the "?" I suggest this:

Add an additional key for toggleHelp: The F1 key (keyCode 112). 
It's well known for its  meaning of "Help".
That would be backwards compatible and should work on all usual keyboards.

PR #3570, opened 2024-02-05
add F1 key to toggleHelp for non-english keyboards

gchriz:
Unfortunately the change of #2645 breaks the toggleHelp on keyboards with a non-english layout. The "?" doesn't have keyCode 191 there.

Since it's probably impossible to catch all international keyboard layouts regarding the "?" I suggest this:

Add an additional key for toggleHelp: The F1 key (keyCode 112).
It's well known for its meaning of "Help".
That would be backwards compatible and should work on all usual keyboards.


hakimel:
Looks good, thanks!

Issue #2838, opened 2020-11-08
Remembered vertical index leads to incorrect jumping


lechten:
When “jumping” in reveal.js presentations with vertical slides, remembered state may lead to wrong navigation behavior.  E.g., open that presentation (Firefox on GNU/Linux and Chrome on Windows): https://oer.gitlab.io/emacs-reveal-howto/howto.html

Wait a bit for the presentation to be fully loaded.  Press “n” 13 times to find yourself on slide 14.  Click into the progress bar, at roughly half of the covered bar.  Find yourself on slide 7.  Press “n” once.  I would expect to find myself on slide 8.  Instead, I’m teleported to slide 14.

The cause for this is a [call to `getPreviousVerticalIndex()` in `slide()`](https://github.com/hakimel/reveal.js/blob/master/js/reveal.js#L1213).  A comment above that call reads: “If we were on a vertical stack, remember what vertical index it was on so we can resume at the same position when returning”
In what situation might we return from where to use this behavior?  My presentations in class work better without this.

(Actually, I do not jump via the progress bar but with the [jump plugin](https://github.com/SethosII/reveal.js-jump-plugin).  The effect is the same.  Use “n” to reach slide 9.  Press “7” and “Enter” to jump to slide 7, press “n” once and find yourself back on slide 9 instead of 8.)

I must admit that I do not understand the role of `oy` in [line 1673](https://github.com/hakimel/reveal.js/blob/master/js/reveal.js#L1673).  If that could also be `indexv`, the functions `getPreviousVerticalIndex()` and `setPreviousVerticalIndex()` might be removed entirely.  Why is that the proper vertical index for slides other than the current one?

gchriz:
As far as I understand your issue I had the same problem when navigating around. I hopefully fixed it and created a PR.

PR #3575, opened 2024-02-08
fix navigateNext to start at top of next vertical stack

gchriz:
Some navigation issue regarding remembered vertical positions drove me crazy until I found out some details now.

After navigating left/right between vertical stacks below the top position and then continuing the presentation with Next (space or N) the presentation jumps from the bottom of current stack to the remembered position of the next stack to the right. And NOT to the top position as expected in the normal flow. 
Navigating backwards (shift-space or P) is working correctly: The bottom slide of the next stack to the left is jumped to. 

As far as I understand it the exact same behaviour is described in the older issue #2838.

So I fixed it in this PR. 

(And I learned about npm etc. so that I was able to build reveal.js' "dist" directory.)


gchriz:
Probably the same fix needs to be applied one "else if" block above for config.rtl and so replacing the simple navigateLeft() there.
Since I don't know anything about RTL I didn't change that yet. But of course can do it later.

PR #3580, opened 2024-02-17
Fix missing fragment events in scrollview

bouzidanas:
`fragmentshown` and `fragmenthidden` events are not triggered when fragments appear and dissappear in scrollview. The code added is just the code that is usually executed that results in the dispatch of these events. This code was carefully placed so that these events are not triggered when they shouldnt be like when the slide comes in and the default fragment automatically appears with the slide. The latter is treated like more of a slide change rather than a fragment being shown.


hakimel:
Good catch. This should definitely be addressed.

We already have logic for dispatching these events in the fragments controller. Instead of duplicating that code in the scroll view I think we should always trigger the events from one place (the fragments controller). https://github.com/hakimel/reveal.js/blob/master/js/controllers/fragments.js#L312-L332

It'll take a little refactoring but shouldn't be major. Will take a stab at it.

hakimel:
Was able to fix this by relocating the fragment event dispatchers so no new code was needed. Added tests too. Thanks again for discovering and highlighting the issue!

bouzidanas:
Even better!

The fragment controller is where I got the code from. While I don't fully understand why event dispatching has to be confined to event target's controller, I do understand that changing the placement of code is better than having duplicate code in multiple locations across multiple files.

I thought about messing with the fragment controller but I was worried I might break things without knowing (or without knowing the issues to test for).

Anyways, I appreciate that you took time to find a better solution!

PR #3588, opened 2024-03-14
Fix error when the notes plugin receives a non-string message

NatKarmios:
It seems the React dev tools also trigger `'message'` events, which pollutes the console with errors (about 2-3 every second).

image

This is especially problematic when developing a presentation with `react-scripts start`, as each error causes a full overlay on the speaker view.

image

The error is caused by the attempt to JSON-parse the event body, and is fixed with a try/catch in `onPostMessage()`.


hakimel:
Thanks for fixing!

Issue #3594, opened 2024-03-23
RevealMath.MathJax3 fails with non-global Reveal instances


jokester:
The plugin expects a _global_ `Reveal` variable with methods like `layout()`, which I believe is not the case since the introduction of the new instancializable API.

If reveal.js was initialized with new   `new Reveal(...)` , MathJax3 will fail with error `TypeError: Reveal.layout is not a function` or alike.

PR #3595, opened 2024-03-23
MathJax3: allow non-singleton Reveal instance

jokester:
fixes https://github.com/hakimel/reveal.js/issues/3594


jokester:
This should fix the worst problem I saw , but there may be other problems around plugins and multiple Reveal instances.

The typical pattern I see in the plugins, see below, may lose references to previous instances if `init()` was called multiple times with different instances

```ts
export const APlugin = () => {
	let deck;

        ....
	return {
		id: 'katex',

		init: function (reveal) {
          	deck = reveal;
        ....
}
```

hakimel:
Thanks for fixing!

PR #3602, opened 2024-03-29
codespell: add config + workflow and make it fix some typos it finds

yarikoptic:
Originally spotted in a "derived" project which bundled reveal.js files -- saw `presenation` and decided to act.

I fixed typos only in source files since I have configured overlong lines to be ignored. So it is still there:

```
❯ git grep -l presenation
dist/reveal.esm.js.map
dist/reveal.js.map
```
please advise if I should regenerate those or would be done automagically/later?


hakimel:
Good addition, thanks! I'll rebuild the minified files.

PR #3603, opened 2024-04-01
Re-add question mark for help

lechten:
Over here, the question mark does not toggle the help overlay any more, probably since that commit: https://github.com/hakimel/reveal.js/commit/0072845828d394bf69af1a2ece9b0f07a43270db

This re-adds a check for keyCode 63, which works for me (German keyboard).


hakimel:
Thanks!

0072845828d394bf69af1a2ece9b0f07a43270db fixed an issue where typing ? inside an input field triggered the help overlay as reported in #2645. The main change was moving the ? listener so that it falls behind existing checks for whether the user has an input focused.

That said I don't believe there should be any issue bringing back the check for keyCode 63.

PR #3619, opened 2024-04-30
Fix two if statements returning bool can be simplified

robvdl:


robvdl:
There is no CI? no GitHub actions?

`gulp test` seems to pass locally at least, though with some existing warnings about puppeteer old headless.

hakimel:
I prefer multiple if cases over ! with multiple conditions for readability.

PR #3630, opened 2024-05-28
fix: fix video-background reset when using notes plugin.

zicklag:
This makes the notes plugin check the difference between the states when responding to slide changes, and appropriately sends `next` or `prev` commands instead of `setState` in order to avoid resetting things like video backgrounds.

I needed this in my project so that I could use the notes view to remote control the slides, without really messing up the video background slide by restarting it every time I triggered the next fragment.


zicklag:
Awesome, thanks!

Issue #3634, opened 2024-06-10
Change the background color.


vansh-sharma-2004:
i will background color from black .

vansh-sharma-2004:
@hakimel please assign this issue to me.

hakimel:
https://github.com/hakimel/reveal.js/pull/3635#issuecomment-2157763811

PR #3635, opened 2024-06-10
change of bg

vansh-sharma-2004:
#3634 issue resolved.


hakimel:
Thanks for the suggestion but I don't want to change the bg color.

Issue #3645, opened 2024-06-20
docs: In React -> Reveal.destroy on unmount does not fully destroy


GiffE:
The docs suggestion on how to use Reveal with react seem to have a bug.

From: https://revealjs.com/react/
```tsx
    const deckRef = useRef<Reveal.Api | null>(null); // reference to deck reveal instance

    useEffect(() => {
        // Prevents double initialization in strict mode
        if (deckRef.current) return;

        deckRef.current = new Reveal(deckDivRef.current!, {
            transition: "slide",
            // other config options
        });

        deckRef.current.initialize().then(() => {
            // good place for event handlers and plugin setups
        });

        return () => {
            try {
                if (deckRef.current) {
                    deckRef.current.destroy();
                    deckRef.current = null;
                }
            } catch (e) {
                console.warn("Reveal.js destroy call failed.");
            }
        };
    }, []);
```

This expressly violates the pitfall described in the React documentation:
https://react.dev/learn/synchronizing-with-effects#dont-use-refs-to-prevent-effects-from-firing

This causes reveal to not unmount properly, the observable bug is that the keyboard bindings from a previous instance remain.

React docs suggest that a cancelation be used instead:
https://react.dev/learn/synchronizing-with-effects#fetching-data

Writing in an abort variable seems to fix the issue along with setting the ref only after initialization.
```tsx
    useEffect(() => {
        let abortInitialize = false;
        let deck = new Reveal(deckDivRef.current!, {
            transition: "slide",
            // other config options
        });

        deck.initialize().then(() => {
    
            if (abortInitialize) {
                deck.destroy();
                return;
            }

            // good place for event handlers and plugin setups

            deckRef.current = deck;
        });

        return () => {
            try {
                abortInitialize = true;
                if (deckRef?.current?.isReady()) {
                    deckRef.current.destroy();
                    deckRef.current = null;
                }
            } catch (e) {
                console.warn('Reveal.js destroy call failed.');
            }
        };
    }, []);
```

Related issue:
https://github.com/hakimel/reveal.js/issues/3593

bouzidanas:
Good catch! 

[...]

PR #3696, opened 2024-10-24
Prevent double initialization

EarlMilktea:
Related to #3645 .

This PR prevents `.initialize` from running multiple times.

doubled

↑: Navigation arrow doubly rendered

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -130,11 +130,15 @@ export default function( revealElement, options ) {
 
 		if( !revealElement ) throw 'Unable to find presentation root (<div class="reveal">).';
 
-		initialized = true;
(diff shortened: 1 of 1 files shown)


hakimel:
Preventing double-initialization is a great idea. To be honest, I was surprised this wasn't already in place 😅

Since we have an `initialized` flag I think it's clearer to use that instead of checking DOM references. I've added this in 16ac4b00671a8ac6433091118c4ebd9b00255545

Let me know if I missed anything and I'll be happy to switch to the solution you proposed instead.

EarlMilktea:
@hakimel 

Your suggestion looks good to me.
Let me thank you here for developing/maintaining this great package!

PR #3716, opened 2024-11-11
Add 'controlsOnlyForSpeaker' option

gpotter2:
Hi. Thanks for this great project.

This tiny PR adds a setting 'controlsOnlyForSpeaker' that only displays the controls for the presenter.

### DIFF
--- a/js/config.js
+++ b/js/config.js
@@ -15,7 +15,9 @@ export default {
 	minScale: 0.2,
 	maxScale: 2.0,
 
-	// Display presentation control arrows
+	// Display presentation control arrows.
+	// This can be a boolean (true / false) or 'speaker-only' to only display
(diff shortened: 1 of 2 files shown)


jdeniau (commented):
I'm not related to reveal.js, but I think that this could be a value for `controls`:

```ts
controls: boolean | 'speaker-only';
```

This way, we could implement `'display-only'` of other kind of extension. Moreover, it avoid adding another prop to the API.

hakimel:
I like @jdeniau's suggestion here. Can you update it so that this mode is enabled via `controls: 'speaker-only'`?

gpotter2:
Done, sorry for the delay.

hakimel:
Looks great, thanks!

PR #3758, opened 2025-02-26
Fix font color issue in Firefox

### LINKED ISSUE #3735: White theme on firefox as white font color
When reading the index.html from firefox without running the npm server, the font is white on a white background. It doesn't happen when `npm start` is used and it doesn't happen in edge.

Dustin4444:
Fixes #3735

Adds a devcontainer configuration file for development environment setup.

* Adds `.devcontainer/devcontainer.json` with tasks for testing, building, and launching the project using npm commands.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,7 @@
+{
+  "tasks": {
+    "test": "npm install && npm test",
+    "build": "npm install && npm run build",
+    "launch": "npm install && npm start"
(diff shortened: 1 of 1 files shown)


.devcontainer/devcontainer.json
Dustin4444, line 1: # - [ ]

hakimel:
Thanks but I don't think dev contains are required by enough reveal.js users to merit adding this to the repo.

PR #3767, opened 2025-03-20
Fixes shortcuts in overlay

### LINKED ISSUE #3766: Shortcuts work behind the scenes in overlay
I start using the overlays for links, and experienced a few limitations around keyboarding while you're in an overlay:
- presentation navigation controls continue to work, but happy invisibly in the background; a couple of these controls are useful to have in overlay mode (e.g. space or arrow down in an iframe), hence, I'd say we should block those events while we're in an overlay.
- pausing the presentation still works in an overlay, but it is not visible since the overlay (`z-index: 1000`) sits on top of the pause (`z-index: 100`); either the feature should not work, or it should cover the overlay imo.

tobi-or-not-tobi:
When an overlay is active, there are a few shortcuts that don't work nicely with overlays:
- the presentation navigates behind the overlay when you're using shortcuts (e.g. arrowDown); I've blocked all keydown events on the overlay, other than fullscreen, pause and escape.
- the pause feature works, but is hidden behind the overlay; I've moved the overlay to `z-index: 99` to ensure it's below the pause element.

It is also inconvenient to not have a shortcut for closing the overlay while we're in fullscreen mode; fullscreen mode will close on escape, which is a bit annoying when you want to close the overlay; shortcut "c" has been added to close the overlay. 

Fixes #3766 

If needed, we could make things more configurable. Also, not sure if there's better a way to leverage existing shortcut class (e.g. to get keyboard configurations).


hakimel:
Thanks for reporting this!

Shortcuts working behind the overlay is a bug. However I prefer addressing this in the global keyboard handler so that there's a single place that provides a full overview of all key binding logic. I've addressed this in b3fd27d071b330e16ed5780945572555f6b028b4

I'm split on whether or not the pause overlay should be allowed on top of the overlay. Currently leaning towards not allowing pause while an overlay is open but maybe there's an argument for allowing it?

tobi-or-not-tobi:
Cool, makes sense to handle this elsewhere. I just started to use reveal (awesome work btw!) and indeed wasn't too sure how to handle it properly. 

I'd say either not support it (and then it should not be done in the background) or support it properly. I'm in favour of supporting it to keep it simple for the user; the user might not even be aware he's in an overlay, he just clicks buttons/links.

hakimel:
Okay, let's allow pausing then. Do you want to update this PR or do you want me to take care of it?

tobi-or-not-tobi:
> Okay, let's allow pausing then. Do you want to update this PR or do you want me to take care of it?

Happy to take care of it.

tobi-or-not-tobi:
I'll wait for the other PR to be finished before doing so

hakimel:
> I'll wait for the other PR to be finished before doing so

Thanks! The other change is not in a PR, it's already in master.

tobi-or-not-tobi:
Closing since the pausing overlay will be fixed in new PR

PR #3768, opened 2025-03-21
fix: move pause overlay over other overlay

### LINKED ISSUE #3766: Shortcuts work behind the scenes in overlay

tobi-or-not-tobi:
When in an overlay, the pause overlay is behind the current overlay. With this fix, the pause-overlay moves up to the 1001 z-index and the overlay keycodes continue to work when an overlay is open. 

Fixes #3766


css/reveal.scss
```diff
@@ -1073,7 +1073,7 @@ $controlsArrowAngleActive: 36deg;
 	background: black;
 	visibility: hidden;
 	opacity: 0;
-	z-index: 100;
+	z-index: 10001;
```
hakimel, line 1076: Instead of bumping z-index up higher, can we leave it at 100 and move the overlay down to 99? Thanks!

tobi-or-not-tobi: Yes totally, I had that before. The reason I switch to moving the pause overlay up, was that I assume the overlay function (begin generic) has potentially being customised by users at more places, and therefore a change to the z-index could harm quit a bit. At the same time, when user have overridden the `z-index` of the pause overlay already, this change will not affect them. 

Either way, happy to follow your recommendation.

hakimel: It's a good consideration but let's go ahead and change that anyway. It's such a small change I doubt it will affect many people.

css/reveal.scss
```diff
@@ -1073,7 +1073,7 @@ $controlsArrowAngleActive: 36deg;
 	background: black;
 	visibility: hidden;
 	opacity: 0;
-	z-index: 10001;
+	z-index: 99;
```
hakimel, line 1076: You changed the index of the pause overlay, but that should remain at 100.

The `.r-overlay` should move to 99 so that it appears below the pause screen.

tobi-or-not-tobi: sorry for that, pls review once more.

hakimel:
Looks good, thanks @tobi-or-not-tobi!

hakimel:
This has been released in 5.2.1 🚀 https://github.com/hakimel/reveal.js/releases/tag/5.2.1

RatStar811 (approved):

PR #3776, opened 2025-03-25
Allow users to use meta keys when navigating to preview links

### LINKED ISSUE #3775: Allow to use meta keys when navigating to preview links
Preview links are opened by default in an iframe. This is a great feature if you don't want to leave the screen, but in cases where you do want that occasionally, or for links that are blocked due to Content Security Policy (CSP), it would be convenient to allow for the default browser behaviour of using meta keys when clicking.

Although revealjs allows to manage using the preview link per individual link, it is (at least in my case) pretty convenient to handle all links as preview links by default. Especially the CSP blockers are annonnying in such case.

The meta key is either the command on mac or the ctrl on windows. E.g. Chrome allows for these key combinations on links
- `meta+click`: open url in new _tab_ (unfocused)
- `meta+shift+click`: open url in new _tab_ (focused)
- `shift+click`: open url in new _window_ (focussed)
- `alt+click`: download url

tobi-or-not-tobi:
When a meta key is hold down while clicking a (preview) link, the preview link feature is prevented. 

closes #3775


hakimel:
Thanks! I'm a big cmd+click user so definitely want that to work.

PR #3780, opened 2025-03-25
feat: Distinct hover vs pointer events on overview

### LINKED ISSUE #3779: Distinct hover vs pointer events on overview slides
While navigating the overview slides I'm often using the keyboard. When you have a number of slides, it's very confusing to understand the active slide. This is caused by the mouse being at the same position, the slides keep moving around (as the viewport moves with the activated slide) and the UI for hovered vs active slides is identical. 

I'd therefor prefer: 
- an alternative visual state for hovered slides (more subtle)
- avoid hovering .present slides
- avoid hovering :active slides

tobi-or-not-tobi:
closes #3779


hakimel:
Thanks but closing this in favor of a different fix — will follow up in #3779

PR #3797, opened 2025-05-14
Fix SVG auto-animation scale and animations.

grahnen:
Auto-animate already acts on SVGs, but it does not work very well. The animations "jump", and it only works for some svg objects. This PR allows animation of the `d` attribute on `<path>`, and fixes the jumpiness of SVG animations with what I think is a hack. Specifically, in autoanimate.js, for things with `<svg>` as parents, we multiply the deltas by a factor of 2. 

I have no idea why it works, and this would probably be good to figure out before merging.


linzhxin (approved):

grahnen:
My arbitrary magic scaling only worked sometimes. This new version instead uses screen coordinates converted from element coordinates for the CSS animations. This makes the auto-animations work in SVGs with `viewBox` attributes.

PR #3807, opened 2025-07-11
chore: ignoring node_modules in codespell

manuelarte:
In some projects, when I use reveal.js, I get codespell issues in the `node_modules` folder, I think it's safe to ignore this folder from the analysis.


hakimel:
Makes sense, thanks.

PR #3820, opened 2025-09-30
Update demo.html - use https

hans2103:
This PR will replace http by https for your website in the demo

http://hakim.se/
into
https://hakim.se/


hans2103:
@hakimel Can you take a look at this small PR?

PR #3907, opened 2026-05-28
fix(progress): don't add fragment progress on the last slide

### LINKED ISSUE #1980: Progress is not correctly calculated when there are fragments on the last slide
When fragments are used on the last slide, the progress calculation does not work. It adds the fragment progress to the past slides count, which is assumed to never go above the total count - 1 (in the function [`getProgress()`](https://github.com/hakimel/reveal.js/blob/43eada79901830702bd40dce857831aef8e76759/js/reveal.js#L3508). This means that the progress value is 1 at the beginning of the last slide and > 1 once the fragments are shown, which makes the progress bar wider than the screen. See example here: [https://joelgoop.github.io/reveal-progress-bug](https://joelgoop.github.io/reveal-progress-bug).

Jefsky:
Fixes #1980.

When fragments are used on the last slide, the fragment progress was added to `pastCount` which already equals `totalCount - 1`, making the result > 1 and causing the progress bar to overflow.

Fix by only adding fragment progress when `pastCount < totalCount - 1`, i.e., when not on the last slide. On the last slide, fragments should not increase the progress beyond 100%.

### DIFF
--- a/js/reveal.js
+++ b/js/reveal.js
@@ -2049,8 +2049,12 @@ export default function( revealElement, options ) {
 				// that is made up by its fragments (0-1)
 				let fragmentWeight = 0.9;
 
-				// Add fragment progress to the past slide count
-				pastCount += ( visibleFragments.length / allFragments.length ) * fragmentWeight;
+				// Only add fragment progress when not on the last slide.
+				// On the last slide, fragments should not increase pastCount
+				// beyond totalCount-1, as that would make progress > 1.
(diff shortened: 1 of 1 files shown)


hakimel:
This was a real issue, but it was fixed in [2019 by clamping progress to 1](https://github.com/hakimel/reveal.js/commit/922d791fd9d74caea9d706dbf39250146082ee62). This PR shouldn't be needed anymore, and could cause getProgress() to return NaN for single-slide decks.

PR #3909, opened 2026-06-03
fix: add missing picocolors dependency and fix test script - I would be glad to become your contributor

MatrixNeoKozak:
### Summary of changes

- Added `picocolors` to `devDependencies` in `package.json`.
- Refactored `scripts/test.js` to use `picocolors` instead of extending `String.prototype` with a non-existent `colors` package, which was causing runtime errors.

### Rationale

The `scripts/test.js` file was attempting to call `.red` and `.green` methods on strings, which are typically added by the `colors` package. However, this package was missing from `dependencies`/`devDependencies`, leading to `TypeError` failures during test execution. `picocolors` is a more lightweight, modern, and widely used alternative, already present in other workspaces within this project.

### Technical impact analysis

- Fixes the broken test suite infrastructure.
- Ensures that test execution correctly reports results with proper console coloring.
- Improves the maintainability of the testing script by replacing the side-effect-heavy `colors` package with explicit `picocolors` imports.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -89,6 +89,31 @@
 				"node": ">=6.9.0"
 			}
 		},
+		"node_modules/@emnapi/core": {
+			"version": "1.10.0",
+			"resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.10.0.tgz",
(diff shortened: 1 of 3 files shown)


Mazzika1 (commented):
مم

PR #3916, opened 2026-07-03
fix: upgrade vulnerable direct dependencies

sonukapoor:
CVE Lite CLI (PR #3915) flagged two direct devDependencies with known advisories. This upgrades both to their patched releases.

vite@8.0.8 carries a high severity advisory. CVE Lite validated 8 versions above the current release before landing on 8.0.16 as the lowest known non-vulnerable version.

esbuild@0.28.0 carries a low severity advisory. The fix is a single patch bump to 0.28.1.

Both are devDependencies used only during the build - no runtime impact on reveal.js itself.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
 			"version": "6.0.1",
 			"license": "MIT",
 			"devDependencies": {
-				"esbuild": "^0.28.0",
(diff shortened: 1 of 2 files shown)


hakimel:
These two and all other vulnerable dependencies in master were just updated in master.

PR #3918, opened 2026-07-07
fix: always mute video overlays in speaker notes

duxovni:
To reproduce this bug, you may need to interact with the Speaker Notes window first (eg, click to reset the timer), or else your browser might block autoplaying video in the notes window.


lbesecker195 (approved):
Verified: `Overlay` holds `this.Reveal`, and `isSpeakerNotes()` is part of the public Reveal API. Muting inside the notes window is the right default since the presenter's audio should come from the main window. LGTM.

PR #3929, opened 2026-09-05
Pin GitHub Actions to commit SHAs

lpmi-13:
Update the mutable version tags to commit SHA hashes. Tools like Dependabot and Renovate speak SHA natively now, so there's no reason not to do this, and it eliminates accidental overwrites in pushes to upstream tags as an issue.


lpmi-13:
If it would be helpful, I would gladly add a simple dependabot config for GHA to this PR, or a followup one.

PR #3934, opened 2026-09-08
fix(markdown): allow attribute manipulation of lists

### LINKED ISSUE #3931: Add `<!-- .element class="a-class" -->` to a `<ul>` or `<ol>`
When using this:

```md
- html
- css

<!-- .element class="a-class" -->
```

I'd expect the class to be set on the `ul` element. Instead, it is placed on the last `li` element.

Image

This is a breaking change; I think it was introduced in reveal 6, committed on Feb 25. See the diff: https://github.com/hakimel/reveal.js/commit/3be72d43f79e98d6d25945e7bf57ef63841781ac#diff-01a5b0a3d47ab76eddb43a532d564cd2666869541dbfaa026ba7ce2400bce216R372-R374

I don't know why it got introduced like this.

Can we revert this line?

Image

nicojs:
Allow attribute manipulation of lists using `<!-- .element class="a-class" -->` syntax. Previously (since Reveal.js 6), attribute manipulation was applied to the last element of a list instead of the list itself.

Workaround for when you want the class on the last element:

```md
- html
- <!-- .element class="a-class" -->
  css
```

Fixes #3931

### DIFF
--- a/plugin/markdown/plugin.js
+++ b/plugin/markdown/plugin.js
@@ -370,12 +370,8 @@ const Plugin = () => {
 		}
 
 		if ( element.nodeType === Node.COMMENT_NODE ) {
-		let targetElement = previousElement;
-		if( targetElement && ( targetElement.tagName === 'UL' || targetElement.tagName === 'OL' ) ) {
-			targetElement = targetElement.lastElementChild || targetElement;
-		}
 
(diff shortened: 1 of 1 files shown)


hakimel:
see https://github.com/hakimel/reveal.js/issues/3931#issuecomment-5620063588

